Issue #28, opened 2018-09-27
Fix TravisCI


ziflex:
The problem is that I cannot install ANTLR4 properly in order to generate parser.

ziflex:
@yiss yep. I haven't tried it. 
If it works on CircleCi - I'm fine to use it :)

krishnakarthik1309:
@ziflex, @yiss submitted a PR #35

ziflex:
@krishnakarthik1309  hey, thanks for your help.
It works now, but it seems that Antlr4 does not generate anything. It just renders list of commands. Probably arguments are not being passed.

krishnakarthik1309:
![screenshot from 2018-10-04 21-28-54](https://user-images.githubusercontent.com/12638352/46487020-99d03a00-c81c-11e8-8841-18475fc52ea6.png)
As you can see even when I run the antlr4 command from `$GOPATH/src/github.com/MontFerret/ferret/pkg/parser` it is just giving list of commands. So the problem is not with golang code. Most likely something with `antlr4`(I am not sure). Try to run this in your system and if it is working then tell me which version of antlr4 you are using.

ziflex:
It works fine on macOS. Otherwise I wouldn't be able to generate the parser :)
<img width="682" alt="screenshot 2018-10-04 12 04 22" src="https://user-images.githubusercontent.com/1607148/46487324-ba24d800-c7cd-11e8-9c1e-2143060127ab.png">

krishnakarthik1309:
Can you show the output of `cd pkg/parser`  and then run `antlr4 -Xexact-output-dir -o fql -package fql -visitor -Dlanguage=Go antlr/FqlLexer.g4 antlr/FqlParser.g4`

krishnakarthik1309:
Sry no need understood the problem.. will try to fix it.

krishnakarthik1309:
This PR #37 will fix it by sending arguments to antlr4. You can view current travis-build output [here](https://travis-ci.com/krishnakarthik1309/ferret/jobs/149776767)

ziflex:
@krishnakarthik1309  thank you for your help, finally we can close the issue.

PR #36, opened 2018-10-04
fix travisCI #28

u5surf:
I fix travis-ci and remove to "It should parse none" in TestJsonParse temporary.
because travis-ci end up referring vendor directory which is in your wrong pkg/stdlib/strings/json.go
I also fix json parsing method  when it comes none type value.
After merge this PR. I'll immediately recover the test of this scenario. then it must be in vendor dir correctly.


ziflex:
Sorry, it has already been resolved by other PR.

PR #39, opened 2018-10-04
Add utility function to get random UA string

### LINKED ISSUE #31: Bad user agent
Currently, we generate random user agent for each document.
The problem is that underlying library sometimes gives us user agent string of an old browser, which might be not supported by scraped website.   

That leads to unpredicted results like this: 

screenshot 2018-10-03 12 05 20

We need to narrow down the list of user agents that represent modern browsers.
Maybe we could switch to this library: https://github.com/avct/uasurfer

mb-14:
This should fix #31


ap0:
It would be nice to be able to lock it down to a single user agent if you want, too -- in my use case I don't care about varying user agent, but it can't be a mobile user agent as that changes the rendering of the page I'm trying to scrape. I currently just hardcoded the UA in while pondering the question.

ziflex:
@ap0 this is a good point!

ziflex:
@ap0 take a look at the update issue.

ziflex:
I'm closing it for now since it needs to rethink the entire idea.

Issue #34, opened 2018-10-04
More unit tests


ziflex:
We need more unit tests in ``runtime`` package.

- [x] pkg/runtime/core
- [ ] pkg/runtime/expressions
- [ ] pkg/runtime/expressions/operators
- [ ] pkg/runtime/expressions/literals
- [ ] pkg/runtime/expressions/clauses

esell:
I'm new to this project but have always had an interest in web scraping :). I can give this a go if nobody else has taken it yet...

ziflex:
@esell sure, let me organize the task better in order to make it easier to track the progress.

esell:
@ziflex, so these tests should be browser based? I'm looking at `pkg/runtime/core` right now and using `errrors.go` as a starting point because it appears to be pretty easy and will allow me to get up-to-speed with goconvey. I just hacked a _really_ basic test together that looks like:

```
func TestError(t *testing.T) { 
        Convey("Should match", t, func() {
                msg := "test message"
                cause := errors.New("cause")                 
                e := errors.Errorf("%s: %s", cause.Error(), msg)
  
                ce := core.Error(cause, msg)                
                So(ce, ShouldNotBeNil)
                So(ce.Error(), ShouldEqual, e.Error())
        })
}
```

Should this be something that would require a mock server to test against or is generating the test data inside of the tests and passing that to the functions ok? As you can likely tell, this code base is brand new to me so I'll have to poke around a bit to get an understanding :).

esell:
Ok cool, I think that makes sense haha. I'll put together a handful of tests and submit a WIP PR to see what you think. If it looks good I'll keep at it, if not we can talk then ;)

ziflex:
@esell you can do it with small chunks. it will be easier to track the progress and make changes.
Like, you started covering ``runtime/core``, that could be one PR. Then you can work on ``pkg/runtime/expressions``, that would be another one.
After all PRs you'll get a hacktoberfest t-shirt :)

ConaGo:
Hi, I am new to this project and would like to work on this if there is still need?!

ziflex:
> Hi, I am new to this project and would like to work on this if there is still need?!

Hey, welcome!

Sure, the overall coverage is still pretty low, so any help is appreciated!

solrac97gr:
Hello, I wanna help with the test some consideration or example I should read

solrac97gr:
Hi this is my PullRequest with my Test for Options

[...]

PR #41, opened 2018-10-04
[WIP] add unit tests for runtime/core

esell:
**Currently work in progress**

### DIFF
--- a/pkg/runtime/core/errors_test.go
+++ b/pkg/runtime/core/errors_test.go
@@ -3,7 +3,7 @@ package core_test
 import (
 	"testing"
 
-	"github.com/esell/ferret/pkg/runtime/core"
+	"github.com/MontFerret/ferret/pkg/runtime/core"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/runtime/core/errors_test.go
```diff
@@ -0,0 +1,35 @@
+package core_test
+
+import (
+	"testing"
+
+	"github.com/esell/ferret/pkg/runtime/core"
```
ziflex: You can clone your fork using project's original path. In this case, you won't need to deal with paths.

esell:
@ziflex, thanks for giving this a once over. The `gofmt` thing is odd as my editor runs that on save. I just manually ran the file through `gofmt` and it spit out the same thing...

I'll keep hacking away on the other stuff and make sure it's all formatted before being pushed.

ziflex:
@esell ok, my bad. looks like GitHub displays the code differently, so I thought you didn't format it. 
Now I checked with existing files - same formatting.

PR #55, opened 2018-10-05
Cleanup linting errors

davad:
* Fix naming convention
* Fix increment-decrement errors
* Fix error with how Context is keyed
* Fix var-declaration errors


pkg/runtime/expressions/operators/logical.go
```diff
@@ -15,6 +15,7 @@ type (
 )
 
 var (
+	//revive:disable-next-line var-declaration
```
ziflex: You can safely change ``var`` to ``const``, my bad.

PR #61, opened 2018-10-06
Add Length The Object function

3timeslazy:


ziflex:
@3timeslazy thank you for your time and efforts, unfortunately, you chose the method to implement that I forgot to remove from the list, since it has already been implemented in context of String functions. I'm really sorry for that.

https://github.com/MontFerret/ferret/blob/master/pkg/stdlib/collections/collection.go#L10

ziflex:
But we can meet in the middle - that implementation does not have unit tests ;)

PR #67, opened 2018-10-07
Implement SCREENSHOT function, closes #21

### LINKED ISSUE #21: Add SCREENSHOT function
Add a functions which takes a screenshot of the open page and returns it as binary data.

Pangoraw:
I have implement the screenshot function. As discussed, the signature is :
```
SCREENSHOT(document [, params]) -> binary
SCREENSHOT(url [, params]) -> binary
```
The params have the following structure :
```
params {
  x Float|Int
  y Float|Int
  width Float|Int
  height Float|Int
  format String
  quality Int
}
```
As `cdp` returns a base64 encoded string, I am looking to implement `FROM_BASE64` in a future PR with `PDF`.

### DIFF
--- a/pkg/stdlib/html/blob.go
+++ b/pkg/stdlib/html/blob.go
@@ -1,0 +1,147 @@
+package html
+
+import (
+	"fmt"
+	"context"
+
+	"github.com/MontFerret/ferret/pkg/runtime/values"
+	"github.com/MontFerret/ferret/pkg/stdlib/html/driver/dynamic"
+
+	"github.com/MontFerret/ferret/pkg/runtime/core"
+)
+
+/*
+ * Take a screenshot of the current page.
+ * @param source (Document) - Document.
+ * @param params (Object) - Optional, An object containing the following properties :
+ * 		x (Float|Int) - Optional, X position of the viewport.
+ * 		x (Float|Int) - Optional,Y position of the viewport.
+ * 		width (Float|Int) - Optional, Width of the viewport.
+ * 		height (Float|Int) - Optional, Height of the viewport.
+ * 		format (String) - Optional, Either "jpeg" or "png".
+ * 		quality (Int) - Optional, Quality, in [0, 100], only for jpeg format.
+ * @returns data (Binary) - Returns a base64 encoded string in binary format.
+ */
+func Screenshot(ctx context.Context, args ...core.Value) (core.Value, error) {
+	err := core.ValidateArgs(args, 1, 2)
+	if err != nil {
+		return values.None, err
+	}
+
+	arg1 := args[0]
+
+	err = core.ValidateType(arg1, core.HTMLDocumentType, core.StringType)
+	if err != nil {
+		return values.None, err
+	}
+
+	var doc *dynamic.HTMLDocument
+	var ok bool
+	if arg1.Type() == core.StringType {
+		buf, err := Document(ctx, arg1, values.NewBoolean(true))
+		if err != nil {
+			return values.None, err
+		}
+		doc, ok = buf.(*dynamic.HTMLDocument)
+		defer doc.Close()
+	} else {
+		doc, ok = arg1.(*dynamic.HTMLDocument)
+	}
+
+	if !ok {
+		return values.None, core.Error(core.ErrInvalidType, "expected dynamic document")
+	}
+
+	screenshotParams := &dynamic.ScreenshotArgs{
+		X:       0,
+		Y:       0,
+		Width:   -1,
+		Height:  -1,
+		Format:  "jpeg",
+		Quality: 100,
+	}
+	if len(args) == 2 {
+		arg2 := args[1]
+		err = core.ValidateType(arg2, core.ObjectType)
+		if err != nil {
+			return values.None, err
+		}
+		params, ok := arg2.(*values.Object)
+		if !ok {
+			return values.None, core.Error(core.ErrInvalidType, "expected object")
+		}
+
+		format, found := params.Get("format")
+		if found {
+			err = core.ValidateType(format, core.StringType)
+			if err != nil {
+				return values.None, err
+			}
+			if !dynamic.IsScreenshotFormatValid(format.String()) {
+				return values.None, core.Error(
+					core.ErrInvalidArgument, 
+					fmt.Sprintf("format is not valid, expected jpeg or png, but got %s", format.String()))
+			} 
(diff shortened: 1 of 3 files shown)


pkg/stdlib/html/driver/dynamic/document.go
```diff
@@ -753,3 +753,57 @@ func (doc *HTMLDocument) Navigate(url values.String) error {
...
+type ScreenshotArgs struct {
+	X       float64
+	Y       float64
+	Width   float64
+	Height  float64
+	Format  string
+	Quality int
+}
+
+func (doc *HTMLDocument) CaptureScreenshot(params *ScreenshotArgs) ([]byte, error) {
```
ziflex: Could you change the output value to ``Binary`` from ``values`` package?

pkg/stdlib/html/blob.go
```diff
@@ -0,0 +1,141 @@
...
+			if err != nil {
+				return values.None, err
+			}
+			screenshotParams.Quality = quality.Unwrap().(int)
+		}
+	}
+
+	bin, err := doc.CaptureScreenshot(screenshotParams)
+	if err != nil {
+		return values.NewBinary([]byte{}), err
```
ziflex: Let's return just ``None``.

pkg/stdlib/html/blob.go
```diff
@@ -0,0 +1,141 @@
...
+ * 		width (Float|Int) - Optional, Width of the viewport.
+ * 		height (Float|Int) - Optional, Height of the viewport.
+ * 		format (String) - Optional, Either "jpeg" or "png".
+ * 		quality (Int) - Optional, Quality, in [0, 100], only for jpeg format.
+ * @returns data (Binary) - Returns a base64 encoded string in binary format.
+ */
+func Screenshot(ctx context.Context, args ...core.Value) (core.Value, error) {
+	err := core.ValidateArgs(args, 1, 2)
+	if err != nil {
+		return values.NewBinary([]byte{}), err
```
ziflex: Let's return just ``None``.

pkg/stdlib/html/blob.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return values.None, err
+	}
+
+	var doc *dynamic.HTMLDocument
+	var ok bool
+	if arg1.Type() == core.StringType {
+		buf, err := Document(ctx, arg1, values.NewBoolean(true))
+		if err != nil {
+			return values.NewBinary([]byte{}), err
```
ziflex: Let's return just ``None``.

pkg/stdlib/html/blob.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		doc, ok = buf.(*dynamic.HTMLDocument)
+		defer doc.Close()
+	} else {
+		doc, ok = arg1.(*dynamic.HTMLDocument)
+	}
+
+	if !ok {
+		return values.NewBinary([]byte{}), core.Error(core.ErrInvalidType, "expected dynamic document")
```
ziflex: Let's return just ``None``.

pkg/stdlib/html/driver/dynamic/document.go
```diff
@@ -753,3 +753,57 @@ func (doc *HTMLDocument) Navigate(url values.String) error {
 
 	return doc.WaitForNavigation(5000)
 }
+
+type ScreenshotArgs struct {
+	X       float64
+	Y       float64
+	Width   float64
+	Height  float64
+	Format  string
```
ziflex: Can we create an enum type instead of the magic string? It will improve the readability.
Something like this (you can come up with a better naming):

```go

type ScreenshotFormat string

const (
  ScreenshotFormatPNG ScreenshotFormat = "png"
  ScreenshotFormatJPEG ScreenshotFormat = "jpeg"
)

```

pkg/stdlib/html/driver/dynamic/document.go
```diff
@@ -753,3 +753,57 @@ func (doc *HTMLDocument) Navigate(url values.String) error {
...
+	Quality int
+}
+
+func (doc *HTMLDocument) CaptureScreenshot(params *ScreenshotArgs) ([]byte, error) {
+	ctx := context.Background()
+	metrics, err := doc.client.Page.GetLayoutMetrics(ctx)
+	
+	if params.Format != "jpeg" && params.Format != "png" {
```
ziflex: Once you add enum, let's add a validation function something like ```IsScreenshotFormatValid(format ScreenshotFormat) bool```.

Pangoraw:
I just updated and squashed :+1: Thanks for the review !

pkg/stdlib/html/blob.go
```diff
@@ -0,0 +1,145 @@
...
+		format, found := params.Get("format")
+		if found {
+			err = core.ValidateType(format, core.StringType)
+			if err != nil {
+				return values.None, err
+			}
+			if !dynamic.IsScreenshotFormatValid(format.String()) {
+				screenshotParams.Format = dynamic.ScreenshotFormatJPEG
+			} else {
+				screenshotParams.Format = dynamic.ScreenshotFormat(format.String())
```
ziflex: ``dynamic.ScreenshotFormat(format.String())`` may return an unpredictable result. What if a format is ``foobar`` string? You'll get a unsupported format.
Let's just return an error saying that the given format is not supported (or invalid)

ziflex: @Pangoraw my bad, you are right! 
But still, I think if a given format is invalid, it's better to inform a user about it.
It's better to be strict in order to avoid side affects.

pkg/stdlib/html/blob.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+
+		format, found := params.Get("format")
+		if found {
+			err = core.ValidateType(format, core.StringType)
+			if err != nil {
+				return values.None, err
+			}
+			if !dynamic.IsScreenshotFormatValid(format.String()) {
+				screenshotParams.Format = dynamic.ScreenshotFormatJPEG
```
ziflex: Basically, instead of the fallback value, let's return an error.

pkg/stdlib/html/blob.go
Pangoraw: You are right ! It's fixed now. Thanks for taking the time to review :+1:

ziflex:
@Pangoraw thank you for your time and efforts!

Issue #10, opened 2018-09-18
Add object functions


ziflex:
- [x] KEYS(object, sort) → strArray
- [x]  HAS(object, keyName) → isPresent
~~LENGTH(object) → count~~ ([Implemented here](https://github.com/MontFerret/ferret/blob/master/pkg/stdlib/collections/length.go#L10))
- [x]  MERGE(object1, object2, ... objectN) → newMergedObject
- [x]  MERGE_RECURSIVE(object1, object2, ... objectN) → newMergedObject
- [x]  VALUES(document, removeInternal) → anyArray
- [x]  ZIP(keys, values) → newObj
- [x] KEEP(object, key1, key2, ... key) → newObj

3timeslazy:
I wanted to work on this issue. If there is any useful additional information - I will be glad to know it.

ziflex:
@3timeslazy great!

1. Read [the documentation](https://docs.arangodb.com/3.4/AQL/Functions/Document.html)
2. Do not implement those methods that are related to ArangoDB Collections like ``IS_SAME_COLLECTION``, ``PARSE_IDENTIFIER`` and etc.
3. Do not mutate values itself - return only new instances.
4. Here is [an example](https://github.com/MontFerret/ferret/tree/master/pkg/stdlib/strings) how to create function. I will merge my work on Array functions, that will be even better reference.

3timeslazy:
I understand correctly that function KEYS the same with ATTRIBUTES function from the doc? 
https://docs.arangodb.com/3.4/AQL/Functions/Document.html#attributes
First of all, Im interesting in the sort argument. Does it have the same purpose?

ziflex:
@3timeslazy you are right. Pretty much the same.

3timeslazy:
link inside https://github.com/MontFerret/ferret/issues/10#issue-361404327 is no longer valid.

ziflex:
@3timeslazy updated.

3timeslazy:
Question about the MERGE function. The documentation supports 2 function signatures: 
1. MERGE(object1, object2, ... objectN) → mergedObject
2. MERGE(objArray) → mergedObject

 In this case need to implement only one (1th signature) or both?

ziflex:
@3timeslazy let's implement both.
Just do a check for a type when you deal with an array.
Which means if any of an array elements is not an object - return ``None``.

Here is ArangoDB's behavior.
[...]

PR #69, opened 2018-10-07
Added objects 'Has' and 'Keys' function

3timeslazy:


pkg/stdlib/objects/has.go
```diff
@@ -0,0 +1,40 @@
...
+	err = core.ValidateType(args[1], core.StringType)
+
+	if err != nil {
+		return values.None, err
+	}
+
+	obj := args[0].(*values.Object)
+	keyName := args[1].(values.String)
+
+	_, has := obj.Get(keyName)
```
ziflex: Let's change return value of ``Object.Get`` to ``(core.Value, values.Boolean)`` for consistency.

3timeslazy:
@ziflex The Keys method got into a commit. Remove it from the commit and add it as other PR along with the tests or leave it as is?

ziflex:
@3timeslazy leave it as it is, just add unit tests, please :)

pkg/stdlib/objects/has.go
```diff
@@ -0,0 +1,40 @@
...
+
+	return values.NewBoolean(bool(has)), nil
```
ziflex: After your change in ``Object`` struct, you can just return ``has`` variable now :)

3timeslazy: I just used 3th statement
https://github.com/MontFerret/ferret/issues/10#issuecomment-427376533

PR #70, opened 2018-10-07
change Object.Get(key values.String) (core.Value, bool) to Object.Get…

3timeslazy:
…(key values.String) (core.Value, values.Boolean)  for consistency.


ziflex:
Awesome, thanks!

PR #78, opened 2018-10-08
add unit tests for runtime/core - helpers

esell:
** NOTE ** there are currently a few tests that are not passing (commented in code). need to dig a bit deeper and see what is going on...

### DIFF
--- a/pkg/runtime/core/helpers.go
+++ b/pkg/runtime/core/helpers.go
@@ -11,12 +11,13 @@ func IsNil(input interface{}) bool {
 		reflect.Array,
 		reflect.Slice,
 		reflect.Map,
-		reflect.Struct,
 		reflect.Func,
 		reflect.Interface,
-		reflect.Chan,
-		reflect.UnsafePointer:
+		reflect.Chan:
 		return val.IsNil()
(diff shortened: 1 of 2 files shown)


esell:
@ziflex, would be interested to get your opinion on the strategy used in these tests as well as your thoughts on why some are failing (returning `nil`).

ziflex:
@esell I guess those cases are not covered in the utility function.

ziflex:
@esell here is an updated version of helper:

```go
func IsNil(input interface{}) bool {
	val := reflect.ValueOf(input)
	kind := val.Kind()

	switch kind {
	case reflect.Ptr,
		reflect.Array,
		reflect.Slice,
		reflect.Map,
		reflect.Func,
		reflect.Interface,
		reflect.Chan:
		return val.IsNil()
	case reflect.Struct,
		reflect.UnsafePointer:
		return false
	case reflect.Invalid:
		return true
	default:
		return false
	}
}
```

If I missed something else - just add.

esell:
I took a look at https://golang.org/pkg/reflect/#Value.IsNil and it appears you got everything so i've updated the PR with those changes.

PR #81, opened 2018-10-08
Adds function to type value into input

ap0:
This adds a function using the CDP Input.DispatchKeyEvent to type an input into a form vs. just setting the value via selector.

### DIFF
--- a/pkg/html/dynamic/document.go
+++ b/pkg/html/dynamic/document.go
@@ -529,7 +529,7 @@ func (doc *HTMLDocument) ClickBySelectorAll(selector values.String) (values.Bool
 	return values.False, nil
 }
 
-func (doc *HTMLDocument) TypeInput(selector values.String, value core.Value) (values.Boolean, error) {
+func (doc *HTMLDocument) InputBySelector(selector values.String, value core.Value, delay values.Int) (values.Boolean, error) {
 
 	valStr := value.String()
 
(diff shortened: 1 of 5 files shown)


pkg/html/dynamic/document.go
```diff
@@ -527,6 +529,49 @@ func (doc *HTMLDocument) ClickBySelectorAll(selector values.String) (values.Bool
...
+
+	if err != nil {
+		return values.False, err
+	}
+
+	if res.Type() == core.BooleanType && res.(values.Boolean) == values.False {
+		return values.False, nil
+	}
+
+	inputClient := input.NewClient(doc.conn)
```
ziflex: Why are you creating a new client ? Document already has one.

ap0: Didn't even notice it.

ap0:
Ok, made the changes as requested to the best of my understanding. `INPUT` handles either HTMLElement or HTMLDocument, and adds the delay value.

pkg/html/dynamic/document.go
```diff
@@ -556,11 +551,25 @@ func (doc *HTMLDocument) InputBySelector(selector values.String, value core.Valu
...
-	if res.Type() == core.BooleanType {
-		return res.(values.Boolean), nil
+	if res.Type() == core.BooleanType && res.(values.Boolean) == values.False {
+		return values.False, nil
 	}
 
-	return values.False, nil
+	delayMs := time.Duration(delay)
+
+	time.Sleep(delayMs * time.Millisecond)
```
ziflex: Why do you need Sleep here?

ziflex: Oh, is it imitation of a pause before starting to type?

pkg/stdlib/html/input.go
```diff
@@ -2,71 +2,100 @@ package html
...
 	}
 
-	// TYPE(el, "foobar")
-	if len(args) == 2 {
+	switch args[0].(type) {
+	case *dynamic.HTMLDocument:
 		arg1 := args[0]
 
-		err := core.ValidateType(arg1, core.HTMLElementType)
+		err = core.ValidateType(arg1, core.HTMLDocumentType)
```
ziflex: You already casted to a target type. This check is redundant.
Let's move you to line 25 and do like this
```go
err = core.ValidateType(arg1, core.HTMLDocumentType, core.HTMLElementType)
```
It will cover both types.

pkg/stdlib/html/input.go
```diff
@@ -2,71 +2,100 @@ package html
...
-		return values.False, err
-	}
+		return doc.InputBySelector(arg2.(values.String), args[2], delay)
 
-	arg2 := args[1]
+	case *dynamic.HTMLElement:
+		arg1 := args[0]
 
-	err = core.ValidateType(arg2, core.StringType)
+		err := core.ValidateType(arg1, core.HTMLElementType)
```
ziflex: Same as above - redundant check.

ziflex:
@ap0 regarding context - where exactly?

pkg/html/dynamic/document.go
ap0: Yes, simulating a delay between clicking and typing.

ap0:
@ziflex re: context, https://github.com/MontFerret/ferret/pull/81/files#diff-a5578584c01b5cd096d8d591fef35429R565

`if err := doc.client.Input.DispatchKeyEvent(nil, ke); err != nil {`

Currently I'm passing `nil`; I wasn't sure if there was some context I should be passing in or using a new one.

ziflex:
Oh, let's pass a ``context.Background()`` for now. It's safer than passing nil.

ap0:
OK, tested with `examples/input.fql` (also added the delay to see it) as well as using the function with both element and document in the scraper I'm writing.

ziflex:
Great! Thank you!

Issue #74, opened 2018-10-08
Improve UA generation


ziflex:
**Is your feature request related to a problem? Please describe.**
The current algorithm of UA generation gives us UA strings of very old browsers which lead to unpredicted results like wrong CSS selectors or inability to render a page at all.

**Describe the solution you'd like**
The solution we need is a list (ideally a smart generator) of modern browsers which has a wide variety of:
- versions (modern ones)
- browser types
- platforms

Libraries to consider:
 - https://github.com/jiusanzhou/surferua
 - https://github.com/malisit/kolpa
 - https://github.com/lorien/user_agent

tvytlx:
Maybe this package is suitable too, https://github.com/EDDYCJY/fake-useragent

can i implement it with this package?

ziflex:
@tvytlx hey, it seems it's more than UA generator and it writes something to FS which I would **not** like to do.
```
If used for the first time, useragent will collect data and create a file in the temporary directory as a file cache, please wait a few seconds
```

tvytlx:
I tried this one https://github.com/malisit/kolpa

here is some random output

```
Opera/9.64.(X11; Linux i686; as-IN) Presto/2.9.182 Version/12.00
(X11; Linux x86_64; rv:1.9.7.20) Gecko/2012-06-05 15:48:37  Firefox/3.6.19.0
Mozilla/5.0 (Windows; U; Windows NT 5.0) AppleWebKit/532.37.6 (KHTML, like Gecko) Version/5.1 Safari/532.37.6
(Macintosh; U; Intel Mac OS X 10_8_0; rv:1.9.3.20) Gecko/2015-02-14 14:02:53  Firefox/3.6.4.0
(Windows NT 5.2) AppleWebKit/5341 (KHTML, like Gecko) Chrome/13.0.851.0 Safari/5341
(X11; Linux x86_64) AppleWebKit/5341 (KHTML, like Gecko) Chrome/13.0.825.0 Safari/5341
(Windows NT 5.1) AppleWebKit/5351 (KHTML, like Gecko) Chrome/13.0.806.0 Safari/5351
Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 5.0; Trident/5.0)
Mozilla/5.0 (compatible; MSIE 6.0; Windows NT 6.2; Trident/3.1)
[...]

PR #83, opened 2018-10-09
feat: improve User-Agent generation

tvytlx:
the sample UA outputs is here https://github.com/MontFerret/ferret/issues/74#issuecomment-427866259

### DIFF
--- a/pkg/html/common/ua.go
+++ b/pkg/html/common/ua.go
@@ -2,18 +2,15 @@ package common
 
 import (
 	"github.com/MontFerret/ferret/pkg/runtime/env"
-	"github.com/corpix/uarand"
+	"github.com/malisit/kolpa"
 )
 
-func GetUserAgent(val string) string {
-	if val == "" {
-		return val
-	}
+var k = kolpa.C()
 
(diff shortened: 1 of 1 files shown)


tvytlx:
I'm not sure if it meets the 'new browsers' requirement, but all the UA template is separate and its version number is hard-code which means you can modify it as you want.

```
Chrome/{{%numericRandomizer 1 13 16%}}.0.{{%numericRandomizer 1 800 900%}}.0

which will generate:

Chrome/[1, 13].0.[800, 900].0    ([x, y] means any integer between x and y)
```

FYI:

https://github.com/malisit/kolpa/blob/master/data/en_US/user_agent_chrome#L1
https://github.com/malisit/kolpa/blob/10c4658a62cbd37e83531a3a2732c1dab63eb6c9/utils.go#L261

ziflex:
@tvytlx sorry for such a late response.
Is there a way how we can narrow down a list of versions? Like, let's use Chrome starting from version X and FireFox from version Y. 
Otherwise, we may end up with the same situation.

PR #87, opened 2018-10-09
Update embedded.go

dmadhushika:


examples/embedded.go
```diff
@@ -1,14 +1,14 @@
...
+import(
 	"context"
 	"encoding/json"
 	"fmt"
 	"github.com/MontFerret/ferret/pkg/compiler"
 	"os"
 )
-
-type Topic struct {
+//struct
```
bcho: This comment looks unnecessary.

ap0:
This seems unnecessary; `gofmt` will keep the spaces between `struct/var` and the `{/(`

ziflex:
The change does not bring any value.

PR #90, opened 2018-10-09
Adds WAIT_ELEMENT to sample as sometimes WAIT_NAVIGATION isn't sufficient

ap0:
The page will often return an empty array because the search results haven't loaded although the navigation has occurred.


ziflex:
Thank you for your PR!

PR #93, opened 2018-10-10
Bugfix/#89 compare

3timeslazy:


3timeslazy:
@ziflex you are right, double iteration was a mistake, so I removed it. Also added quick check.

PR #94, opened 2018-10-10
add unit tests for runtime/core - scope

esell:

### DIFF
--- a/pkg/runtime/core/scope_test.go
+++ b/pkg/runtime/core/scope_test.go
@@ -1,0 +1,30 @@
+package core_test
+
+import (
+	"testing"
+
+	"github.com/MontFerret/ferret/pkg/runtime/core"
+	"github.com/MontFerret/ferret/pkg/runtime/values"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestScope(t *testing.T) {
+	Convey("Should match", t, func() {
+		rs, cf := core.NewRootScope()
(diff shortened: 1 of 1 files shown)


esell:
@ziflex, can you give the latest commit a review and see if that tests the scenario you were looking for? I just forked an existing scope that had a set variable and then ran some tests against the forked Scope.

pkg/runtime/core/scope_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"testing"
+
+	"github.com/MontFerret/ferret/pkg/runtime/core"
+	"github.com/MontFerret/ferret/pkg/runtime/values"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestScope(t *testing.T) {
+	Convey("Should match", t, func() {
+		rs, cf := core.NewRootScope()
```
ziflex: Looks good.
Let's the final change.

RootScope - varA
Child1Scope - varB
Child2Scope - empty

Add a check that Child2Scope is able to get a variable from the root scope. (it will test functionality of scope traversing)

ziflex: You can just add as a 2nd Convey suit.

esell: I think the new `TestScopeTraversing` should cover this situation, let me know your thoughts :)

PR #96, opened 2018-10-10
Add mutex and check if closed prior to closing

ap0:
I thought my previous PR fixed this problem, but it seems that the callback can get called more than once, causing a panic since the channel was already closed on a previous invocation. I haven't been able to figure out how/why, but it shouldn't cause the app to panic, so this is a stopgap until we can figure out what's going on.

### DIFF
--- a/pkg/html/dynamic/document.go
+++ b/pkg/html/dynamic/document.go
@@ -558,8 +558,15 @@ func (doc *HTMLDocument) WaitForClassAll(selector, class values.String, timeout 
 
 func (doc *HTMLDocument) WaitForNavigation(timeout values.Int) error {
 	onEvent := make(chan struct{})
+	closed := false
+	var mu sync.Mutex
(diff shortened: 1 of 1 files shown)


ziflex:
@ap0 hmmm. might it be because the listener was not removed? 
I'm planning to rewrite that EventBroker, since it has lots of locks, not idiomatic and still have some concurrency issues.

ziflex:
And yeah, it feels dirty :) We need to figure out why it gets called after ``RemoveEventListener`` was called.

ziflex:
@ap0 could you create an issue for this bug? Then we could keep track of it.
Additionally, you may take a look at [this branch](https://github.com/MontFerret/ferret/tree/refactoring/new-event-broker) with ongoing work on new event broker. It does't work though :)
The idea is to simplify it by making it less generic and more straightforward and eliminate chances of race conditions.

PR #101, opened 2018-10-11
Feature/#95 deepclone

3timeslazy:


pkg/runtime/core/cloneable.go
```diff
@@ -0,0 +1,16 @@
...
+	ArrayType:  true,
+	ObjectType: true,
+}
+
+type Cloneable interface {
+	Value
+	Clone() Cloneable
+}
+
+func IsCloneable(value Value) bool {
```
ziflex: Could you move this helper function to ``values`` package? It's very implementation sensitive code.

pkg/runtime/values/object.go
```diff
@@ -202,3 +202,20 @@ func (t *Object) Remove(key String) {
 func (t *Object) SetIn(path []core.Value, value core.Value) error {
 	return SetIn(t, path, value)
 }
+
+func (t *Object) Clone() core.Cloneable {
+	cloned := NewObject()
+
+	var value core.Value
+	var keyString String
+	for _, key := range t.Keys() {
```
ziflex: Let's use ``t.value`` instead. Now, we get an extra iteration over inner map.

3timeslazy:
@ziflex done :)

pkg/runtime/core/cloneable.go
3timeslazy: What about make```IsCloneable(value values.Value) values.Boolean```?

ziflex: @3timeslazy I do not have a strong opinion on this since it's for internal use only. 
It's up to you :)

PR #111, opened 2018-10-13
added MERGE object function

3timeslazy:


3timeslazy:
Too many commits, so it's best to just watch ```Files changed```

pkg/stdlib/objects/merge.go
```diff
@@ -0,0 +1,67 @@
...
+func mergeArray(arr *values.Array) *values.Object {
+	merged := values.NewObject()
+
+	var val core.Value
+	var k values.String
+	var obj *values.Object
+
+	for idx := values.NewInt(0); idx < arr.Length(); idx++ {
+		obj = arr.Get(idx).(*values.Object)
+		for _, key := range obj.Keys() {
```
ziflex: Why don't you use ``Object.ForEach`` instead? You would eliminate 1 extra iteration at least. (by removing ``.Keys`` method)

pkg/stdlib/objects/merge.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func mergeArray(arr *values.Array) *values.Object {
+	merged := values.NewObject()
+
+	var val core.Value
+	var k values.String
+	var obj *values.Object
+
+	for idx := values.NewInt(0); idx < arr.Length(); idx++ {
```
ziflex: Wouldn't ``Array.ForEach`` be more easy to use and read?

pkg/stdlib/objects/merge.go
```diff
@@ -0,0 +1,67 @@
...
+		obj = arr.Get(idx).(*values.Object)
+		for _, key := range obj.Keys() {
+			k = values.NewString(key)
+			val, _ = obj.Get(k)
+			if values.IsCloneable(val) {
+				val = val.(core.Cloneable).Clone()
```
ziflex: Do you think we really need to clone it? I was thinking of just copying.
Could you elaborate on that?

3timeslazy: Variables in AQL are immutable once assigned. So, if we simply apply ```Copy```, then we will break this principle and we will be able to change the nested objects from the outside.

pkg/stdlib/objects/merge.go
3timeslazy: Always forget about this feature :) will fix it

pkg/stdlib/objects/merge.go
3timeslazy: So, do I continue to use the ```Clone``` function?

ziflex: Yep.

PR #115, opened 2018-10-13
Implement FROM_BASE64 function

Pangoraw:
I implemented a `FROM_BASE64` function which can be used with the `SCREENSHOT` and `PDF` functions. So we can just run from
`query.fql`
```aql
RETURN FROM_BASE64(SCREENSHOT('https://google.com'))
```
```bash
ferret query.fql > google.jpeg
```
Instead of using another tool to convert the base64 string.

### DIFF
--- a/pkg/stdlib/strings/decode.go
+++ b/pkg/stdlib/strings/decode.go
@@ -1,0 +1,31 @@
+package strings
+
+import (
+	"context"
+	"encoding/base64"
+
+	"github.com/MontFerret/ferret/pkg/runtime/values"
+
+	"github.com/MontFerret/ferret/pkg/runtime/core"
+)
+
+/*
+ * Returns the value of a base64 representation.
+ * @param base64String (String) - The string to decode.
+ * @returns value (String) - The decoded string.
+ */
+func FromBase64(_ context.Context, args ...core.Value) (core.Value, error) {
+	err := core.ValidateArgs(args, 1, 1)
+	if err != nil {
+		return values.EmptyString, err
+	}
+
+	value := args[0].String()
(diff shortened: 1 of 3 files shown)


pkg/stdlib/strings/decode.go
```diff
@@ -0,0 +1,35 @@
...
+ * @returns value (String) - The decoded string.
+ */
+func FromBase64(_ context.Context, args ...core.Value) (core.Value, error) {
+	err := core.ValidateArgs(args, 1, 1)
+	if err != nil {
+		return values.EmptyString, err
+	}
+
+	value := args[0].String()
+	fmt.Printf("%s", value)
```
ziflex: I think you forgot to clean up something :)

pkg/stdlib/strings/decode.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+	"context"
+	"encoding/base64"
+
+	"github.com/MontFerret/ferret/pkg/runtime/values"
+
+	"github.com/MontFerret/ferret/pkg/runtime/core"
+)
+
+// TODO: Investigate error exit code
```
ziflex: What error code?

Pangoraw:
This was because, I wasn't able to fully test because the test string had some padding to it `Zm9vYmFy==` which would return something like `foobar\n` instead of `foobar`. Sorry for not cleaning it tough :+1:

ziflex:
@Pangoraw I see. The reason of that is that output is returned as a JSON string.

ziflex:
Thank you for your help!

PR #128, opened 2018-10-15
Feature/#10 values

3timeslazy:
There is bug into PR. In 1 case out of 10 the ```When complex type attributes (object)```  test falls:
screenshot 2018-10-16 at 01 11 00
Need help to fix it before merge.

### DIFF
--- a/pkg/runtime/values/array.go
+++ b/pkg/runtime/values/array.go
@@ -228,7 +228,7 @@ func (t *Array) Sort() *Array {
 	copy(c, t.value)
 
 	sort.SliceStable(c, func(i, j int) bool {
-		return c[i].Compare(c[j]) == 0
+		return c[i].Compare(c[j]) == -1
 	})
 
 	res := new(Array)
--- a/pkg/stdlib/objects/lib.go
+++ b/pkg/stdlib/objects/lib.go
@@ -4,10 +4,11 @@ import "github.com/MontFerret/ferret/pkg/runtime/core"
 
 func NewLib() map[string]core.Function {
 	return map[string]core.Function{
-		"HAS":   Has,
-		"KEYS":  Keys,
-		"KEEP":  Keep,
-		"MERGE": Merge,
-		"ZIP":   Zip,
+		"HAS":    Has,
+		"KEYS":   Keys,
+		"KEEP":   Keep,
+		"MERGE":  Merge,
+		"ZIP":    Zip,
+		"VALUES": Values,
 	}
 }
--- a/pkg/stdlib/objects/values.go
+++ b/pkg/stdlib/objects/values.go
@@ -3,6 +3,8 @@ package objects
 import (
 	"context"
 
+	"github.com/MontFerret/ferret/pkg/runtime/values"
+
 	"github.com/MontFerret/ferret/pkg/runtime/core"
 )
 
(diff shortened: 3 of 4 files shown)


3timeslazy:
As I understand it, the problem is the ```Array.Sort``` function.

ziflex:
@3timeslazy all tests passed.

ziflex:
Oh I didn’t notice “1 out of 10”
I will look into that.

ziflex:
@3timeslazy i think your problem is in your unit tests.
Go does not guarantee same order during iterations over maps.  Which means, you should compare result arrays by their items without their order.

3timeslazy:
@ziflex but I sorted both arrays. 
screenshot 2018-10-16 at 21 46 41

3timeslazy:
@ziflex also we can see on the screenshot in the description that ```expected``` array (sorted initially) was reversed by ```Sort```

ziflex:
@3timeslazy did you try to debug it?

3timeslazy:
@ziflex yep. Also I ran ```Sort``` in a cycle 100 and 1000 times but everything was OK. It falls only on this test.

ziflex:
That's weird.
There is a bug either in one of the ``Compare`` methods or... ``StableSort``? :)

ziflex:
@3timeslazy can you add a length comparison before the last assertion?

3timeslazy:
@ziflex did you mean this? 
```go
So(err, ShouldBeNil)
So(actualSorted.Length().Compare(expected.Length()), ShouldEqual, 0)
So(actualSorted.Compare(expected), ShouldEqual, 0)
```

ziflex:
Yep

ziflex:
@3timeslazy what if you do a stress test with sort-compare unit test? with arrays of objects.

ziflex:
> @ziflex yep. Also I ran Sort in a cycle 100 and 1000 times but everything was OK. It falls only on this test.

Same as you did there.
Just try to do many iterations of these operations.

ziflex:
Btw, you can join our chat where we can discuss it

https://discord.gg/kzet32U

pkg/runtime/values/object.go
```diff
@@ -79,15 +79,24 @@ func (t *Object) Compare(other core.Value) int {
 		sortedOther := sort.StringSlice(other.Keys())
 		sortedOther.Sort()
 
+		var tKey, otherKey String
+		var tVal, otherVal core.Value
+
 		for i := 0; i < len(t.value) && res == 0; i++ {
-			if sortedT[i] == sortedOther[i] {
-				continue
```
ziflex: I thought you would put value comparison here.

pkg/runtime/values/object.go
```diff
@@ -73,18 +76,29 @@ func (t *Object) Compare(other core.Value) int {
...
-			if val, exists = t.value[key]; exists {
-				res = val.Compare(otherVal)
+			if sortedT[i] < sortedOther[i] {
+				res = 1
+				break
 			}
 
-			return res == 0
-		})
+			res = -1
```
ziflex: Break? It does not make sense to compare next values if ``res`` != 0

pkg/runtime/values/object.go
```diff
@@ -73,18 +76,29 @@ func (t *Object) Compare(other core.Value) int {
...
+		sortedOther := sort.StringSlice(other.Keys())
+		sortedOther.Sort()
+
+		var tVal, otherVal core.Value
+
+		for i := 0; i < len(t.value) && res == 0; i++ {
+			if sortedT[i] == sortedOther[i] {
+				tVal, _ = t.Get(NewString(sortedT[i]))
+				otherVal, _ = other.Get(NewString(sortedOther[i]))
+				res = tVal.Compare(otherVal)
```
ziflex: Let's compare the ``res``. If it does not equal to zero, we need to stop the iteration.

3timeslazy: Done

Issue #14, opened 2018-09-18
Add DOWNLOAD function


ziflex:
Add ``DOWNLOAD(content) -> binary`` function which would download binary data.

Pangoraw:
Hi,

I am thinking about how to implement this feature. The way I see it, is to use chromium to download to a temporary folder (using [this](https://godoc.org/github.com/mafredri/cdp/protocol/page#SetDownloadBehaviorArgs)). And when the download is done, return the binary data from the downloaded file then remove the temporary folder.

Do you see another way ?

ziflex:
@Pangoraw I was thinking of using just plain HTTP client for that.

Pangoraw:
For static document yes, but if you want to scrap documents in dynamic documents which requires authentification (for example), this could not be enough ?

ziflex:
#120
As a reference

PR #132, opened 2018-10-16
Implement DOWNLOAD function

Pangoraw:
As discussed in #14, I have implement only the signature :
```aql
DOWNLOAD(url)
```
Because, as of right now, I think there is no way to detect whether or not a download is done or not with cdp. We should keep an eye on this [issue](https://github.com/GoogleChrome/puppeteer/issues/3360).


ziflex:
Thank you for the PR!

PR #140, opened 2018-10-24
Feature/#10 merge recursive

3timeslazy:


pkg/stdlib/objects/merge_recursive.go
```diff
@@ -0,0 +1,66 @@
...
+	merged := values.NewObject()
+
+	for _, arg := range args {
+		merged = merge(merged, arg.(*values.Object)).(*values.Object)
+	}
+
+	return merged.Clone(), nil
+}
+
+func merge(src, dst core.Value) core.Value {
```
ziflex: If it's a private method and you use it for merging objects, why don't you define its arguments and the output as the``*values.Object`` type?

3timeslazy: because this function can take values of different types at runtime

pkg/stdlib/objects/merge_recursive.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, arg := range args {
+		if err = core.ValidateType(arg, core.ObjectType); err != nil {
+			return values.None, err
+		}
+	}
+
+	merged := values.NewObject()
+
+	for _, arg := range args {
+		merged = merge(merged, arg.(*values.Object)).(*values.Object)
```
ziflex: Ok, do we really need to cast ``arg`` to ``*values.Object`` if ``merge`` accepts ``core.Value``?

3timeslazy: Nope :) My bad

ziflex:
Great! Thank you for your PR.

PR #146, opened 2018-10-27
add unit test for runtime/expressions - return

agneum:
Issue https://github.com/MontFerret/ferret/issues/34


ziflex:
Thank you for your PR!

PR #149, opened 2018-10-28
Bump Go versions and use '.x' to always get latest patch versions

HaraldNordgren:


ziflex:
Oh, that's a good idea!
Thank you for your PR!

PR #151, opened 2018-10-29
Feature/#106 fmt function

3timeslazy:


pkg/stdlib/strings/fmt.go
```diff
@@ -0,0 +1,88 @@
...
+
+	formatted, err := format(args[0].String(), args[1:]...)
+	if err != nil {
+		return values.None, err
+	}
+
+	return values.NewString(formatted), nil
+}
+
+func format(template string, args ...core.Value) (string, error) {
```
ziflex: Could you change ``args ...core.Value`` to ``args []core.Value``?

pkg/stdlib/strings/fmt.go
```diff
@@ -0,0 +1,88 @@
...
+	rgx, err := regexp.Compile("{[0-9]*}")
+	if err != nil {
+		return "", fmt.Errorf("failed to build regexp: %v", err)
```
ziflex: Let's be consistent in terms of errors. Could you use ``https://github.com/pkg/errors`` instead of ``fmt``?

pkg/stdlib/strings/fmt.go
3timeslazy: Sure. But what's the difference?

ziflex:
Great job! Thank you for your time and efforts!

ziflex:
Closes #151

PR #162, opened 2018-11-05
rename utils.LOG -> utils.PRINT

3timeslazy:


pkg/stdlib/utils/lib.go
```diff
@@ -4,7 +4,7 @@ import "github.com/MontFerret/ferret/pkg/runtime/core"
 
 func NewLib() map[string]core.Function {
 	return map[string]core.Function{
-		"WAIT": Wait,
-		"LOG":  Log,
+		"WAIT":  Wait,
+		"PRINT": Log,
```
ziflex: What about the function's name itself? :)

ziflex:
Thank you for the PR!

PR #164, opened 2018-11-06
RIGHT() function should return substr counting from right rather than…

kiinoo:
RIGHT("abc",1) returns "bc".
but the expected result should be "c"?


ziflex:
Thank you for the PR!

PR #165, opened 2018-11-06
added DATE_ADD, DATE_SUBTRACT functions

3timeslazy:


ziflex:
Great! 2 more and we are done with DateTime functions!
Good job and thank you!

PR #181, opened 2018-11-18
fixed a bug when adding/subtrating did not take an amount of units

3timeslazy:


ziflex:
@3timeslazy thank you for the PR!

PR #182, opened 2018-11-19
added DateCompare function

3timeslazy:


ziflex:
Great! We are finally done with datetime functions! 
Thank you for your fantastic work and commitment!

PR #210, opened 2018-12-18
Added ELEMENT_EXISTS function

ziflex:


e2e/main.go
```diff
@@ -48,6 +54,19 @@ func main() {
 		Dir:  filepath.Join(*pagesDir, "dynamic"),
 	})
 
+	var filterR *regexp.Regexp
+
+	if *filter != "" {
+		r, err := regexp.Compile(*filter)
+
+		if err != nil {
```
3timeslazy: I think would be better to use ```regexp.MustCompile(*filter)``` instead.

e2e/main.go
```diff
@@ -3,12 +3,12 @@ package main
 import (
 	"flag"
 	"fmt"
-	"os"
-	"path/filepath"
-
 	"github.com/MontFerret/ferret/e2e/runner"
 	"github.com/MontFerret/ferret/e2e/server"
 	"github.com/rs/zerolog"
+	"os"
```
3timeslazy: Here you can use ```goimports``` util.

pkg/drivers/cdp/element.go
```diff
@@ -679,6 +679,33 @@ func (el *HTMLElement) CountBySelector(selector values.String) values.Int {
...
+			Msg("failed to retrieve nodes by selector")
+
+		return values.False
+	}
+
+	if res.NodeID == 0 {
+		return values.False
+	}
+
+	return values.True
```
3timeslazy: What about ```return values.NewBoolean(res.NodeID == 0)```?

pkg/drivers/http/element.go
```diff
@@ -248,6 +248,16 @@ func (el *HTMLElement) CountBySelector(selector values.String) values.Int {
...
 }
 
+func (el *HTMLElement) ExistsBySelector(selector values.String) values.Boolean {
+	selection := el.selection.Closest(selector.String())
+
+	if selection == nil {
+		return values.False
+	}
+
+	return values.True
```
3timeslazy: The same with [this](https://github.com/MontFerret/ferret/pull/210/files#r242466319).

e2e/main.go
ziflex: That would panic the app without a proper message.

e2e/main.go
ziflex: @3timeslazy Could you give me more details on that?

pkg/drivers/cdp/element.go
ziflex: @3timeslazy what about code readability? :)

e2e/main.go
3timeslazy: https://github.com/bradfitz/goimports

e2e/main.go
3timeslazy: It will.

3timeslazy: It will.

pkg/drivers/cdp/element.go
3timeslazy: ¯\_(ツ)_/¯

PR #213, opened 2018-12-19
Prevent memory leak

Pangoraw:
Not adding `resp.Body.Close()` will cause the TCP packet to leak since `go` expects us to close the query ([source](https://github.com/bradfitz/exp-httpclient/blob/master/problems.md)).


ziflex:
@Pangoraw good catch! Thank you !

PR #222, opened 2019-01-16
Allows simulated clicks to be canceled

jasonparekh:
User-generated mouse clicks are cancelable (and thus preventDefault() works), but the simulated MouseEvents were not cancelable.  This led to some sites to have different behavior between user-generated and simulated clicks.

See https://developer.mozilla.org/en-US/docs/Web/Events/click.

### DIFF
--- a/pkg/drivers/cdp/document.go
+++ b/pkg/drivers/cdp/document.go
@@ -386,7 +386,7 @@ func (doc *HTMLDocument) ClickBySelector(selector values.String) (values.Boolean
 				return false;
 			}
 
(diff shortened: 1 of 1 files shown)


ziflex:
Good catch!
But how does it work with element's [Click](https://github.com/MontFerret/ferret/blob/master/pkg/drivers/cdp/element.go#L739) method?

ziflex:
@jasonparekh ok, could you make changes [here](https://github.com/MontFerret/ferret/blob/master/pkg/drivers/cdp/events/dispatch.go#L19) as well?

PR #223, opened 2019-01-17
Rename KEEP to KEEP_KEYS

3timeslazy:
I found collision between [lexer](https://github.com/MontFerret/ferret/blob/master/pkg/parser/fql/fql_lexer.go#L290) and [object function](https://github.com/MontFerret/ferret/blob/master/pkg/stdlib/objects/lib.go#L9). This is my version of fix :)


ziflex:
Great! Thank you for the PR!

PR #228, opened 2019-02-07
Type system refactoring

ziflex:
- New type of ``core.Type``. Now it's an interface.
- Moved all HTML related types to ``drivers`` package
- Updated standard library


ziflex:
Too many things has been changed.
Needs to split the PR into 2 separate ones.

PR #233, opened 2019-02-13
Move value specific Iterable and Iterator interfaces into core module

ziflex:


pkg/runtime/core/value.go
```diff
@@ -18,6 +18,17 @@ type (
 		Copy() Value
 	}
 
+	// Iterable represents an interface of a value that can be iterated by using an iterator.
+	Iterable interface {
```
3timeslazy: `Cloneable` able to clone itself, `Measurable` able to measure itself. So `Iterable` able to iterate itself. Seems logical.

3timeslazy:
after [this](https://github.com/MontFerret/ferret/pull/233#pullrequestreview-203775078) I would suggest doing this
```golang
switch collection := data.(type) {
	case core.Iterable:
		iterator, err := collection.Iterate(ctx)
		if err != nil {
			return nil, err
		}

		return collections.NewCollectionIterator(ds.valVariable, ds.keyVariable, iterator)
	case collections.KeyedCollection:
		return collections.NewKeyedIterator(ds.valVariable, ds.keyVariable, collection)
	case collections.IndexedCollection:
		return collections.NewIndexedIterator(ds.valVariable, ds.keyVariable, collection)
	default: ...
}
```
It's make line shorted and more readable.

pkg/runtime/core/value.go
ziflex: ``collections.KeyedCollection`` represents key based collection (Object) where ``collections.IndexedCollection`` represents index based collections (Array).
Iteration logic is different.

pkg/runtime/expressions/data_source.go
```diff
@@ -65,16 +64,17 @@ func (ds *DataSource) Iterate(ctx context.Context, scope *core.Scope) (collectio
...
-				return collections.NewIndexedIterator(ds.valVariable, ds.keyVariable, data.(collections.IndexedCollection))
+				return collections.NewIndexedIterator(ds.valVariable, ds.keyVariable, collection)
 			default:
 				return nil, core.TypeError(
 					data.Type(),
 					types.Array,
 					types.Object,
 					types.HTMLDocument,
 					types.HTMLElement,
+					core.NewType("Iterable"),
```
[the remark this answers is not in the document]
ziflex: Anonymous type for a better error message.

PR #235, opened 2019-02-20
Added Context to HTML methods

ziflex:


3timeslazy:
Everything is OK, except for the fallen build.

PR #274, opened 2019-03-27
add go111module

3timeslazy:


.travis.yml
```diff
@@ -45,7 +53,8 @@ jobs:
     - make vet
     - make lint
     - make fmt
-    - if [[ $(git diff --stat) != '' ]]; then echo 'Invalid formatting!' >&2; exit 1; fi
+    - git diff
```
ziflex: Why did you make the change?

3timeslazy: @ziflex to see `git diff` when pipelines fall and fix the problem.

PR #281, opened 2019-04-10
Fixes WaitForNavigation race condition in #125

### LINKED ISSUE #125: WaitForNavigation callback can get called more than once.
**Describe the bug**
The callback in WaitForNavigaton can get called more than once, causing a panic due to closing a channel that is used to signal the page load event closing again.

**To Reproduce**
Not able to reproduce consistently.

**Expected behavior**
Should remove the handler from the page load event and not call the callback more than once.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version 69

**Additional context**
https://github.com/MontFerret/ferret/pull/96

ap0:
When multiple redirects happen very quickly, sometimes the event handler will not be removed before the next page is loaded, so it will call the handler twice and panic because it's attempting to close a previously-closed channel.

Closes #125

### DIFF
--- a/pkg/drivers/cdp/document.go
+++ b/pkg/drivers/cdp/document.go
@@ -707,8 +707,11 @@ func (doc *HTMLDocument) WaitForClassBySelectorAll(ctx context.Context, selector
 
 func (doc *HTMLDocument) WaitForNavigation(ctx context.Context) error {
 	onEvent := make(chan struct{})
+	var once sync.Once
 	listener := func(_ context.Context, _ interface{}) {
-		close(onEvent)
+		once.Do(func() {
+			close(onEvent)
+		})
 	}
 
 	defer doc.events.RemoveEventListener(events.EventLoad, listener)


ziflex:
Good use case. I wonder how stable it is in this case. If many redirects occur, it handles only first one, it misses the rest which might also lead to some unpredictable behavior.

ap0:
Yeah, I don't know enough about how this works to say.  Right now, though, it crashes whenever it encounters this.  Would it be better to somehow have the event broker map events to specific pages when it loads instead?

ziflex:
What if we create a debounce function to handle that?
Something like this:
https://gist.github.com/jonstaryuk/133ced9882aa00502fa7d1db31cb5253

ap0:
I like that solution, but what about the case where it's in the middle of a redirect when the debounce function finally calls the callback (i.e. one of the intermediate pages is slow but will still redirect)?

The big thing for me here is to just guard against calling `close()` on a channel twice, which I don't think that would guarantee, but would make the chances of it happening less likely.

ziflex:
Good point. I think in this case it may just fallback to this initial solution with ``sync.Once``.

ziflex:
Ok, I will merge it.  But we need to improve it in the future.

PR #283, opened 2019-04-12
Added delay randomization for inputs

ziflex:
Added delay randomization for inputs. It allows to bypass bots protection systems.


Makefile
```diff
@@ -13,6 +14,9 @@ default: build
 
 build: vet generate test compile
 
+install:
```
3timeslazy: `export GO111MODULE=on` is enough. Go download dependencies yourself.

PR #326, opened 2019-07-09
Feature/#293 regular exp operator

ziflex:
- ``=~`` positive matching
- ``!~`` negative matching


pkg/compiler/visitor.go
```diff
@@ -1187,6 +1188,38 @@ func (v *visitor) doVisitEqualityOperator(ctx *fql.ExpressionContext, scope *sco
...
+	case literals.StringLiteral:
+		_, err := regexp.Compile(string(lit))
+
+		if err != nil {
+			src := v.getSourceMap(rawExps[1])
+
+			return nil, errors.Wrap(err, src.String())
+		}
+
+		break
```
golangcibot: S1023: redundant break statement (from `gosimple`)

ziflex:
> What about flags e.g.
> 
> * g (global) - don't return after first match
> * m (multiline) - ^ and $ match start/end of line
> * i (insensitive) - case insensitive match
> * u (ungreedy) - make quantifiers lazy
>   ?

It supports everything that is supported by [Regexp.MatchString](https://golang.org/pkg/regexp/#Regexp.MatchString).

PR #333, opened 2019-07-15
Bugfix/#332 nav deadlock

### LINKED ISSUE #332: WAIT_NAVIGATION timeout error ,the newest master version
update to the newest master version
run the embeded script in example directory:
github.com/MontFerret/ferret/examples/embedded.go

```
query := `
		LET doc = DOCUMENT("https://www.baidu.com",true) //any link
		WAIT_NAVIGATION(doc,30000)
		RETURN doc.url
	`
```

the error info is:
`operation timed out: WAIT_NAVIGATION(doc,30000) at 3:2`

I debuged the code and found that :

1.the EventLoad is registed at 2 place ok(page's NewHTMLPage,page's WaitForNavigation)
2.the page load event(EventLoad) is triggered(broker's emit) ok,(by close ch)
3.the page 's NewHTMLPage received the load event ok.
4.but the page's WaitForNavigation does not received this event???

```
func (p *HTMLPage) WaitForNavigation(ctx context.Context) error {
	onEvent := make(chan struct{})
	var once sync.Once
	listener := func(_ context.Context, _ interface{}) {
		once.Do(func() {
			close(onEvent)
		})
	}

	defer p.events.RemoveEventListener(events.EventLoad, listener)

	p.events.AddEventListener(events.EventLoad, listener)  **//<--not received the load event**

	select {
	case <-onEvent:
		return nil
	case <-ctx.Done():
		return core.ErrTimeout
	}
}
```

ziflex:
Fixes #332

### DIFF
--- a/pkg/drivers/common/lazy.go
+++ b/pkg/drivers/common/lazy.go
@@ -9,8 +9,10 @@ import (
 )
 
 type (
+	// LazyValueFactory represents a value initializer
 	LazyValueFactory func(ctx context.Context) (core.Value, error)
 
+	// LazyValue represents a value with late initialization
 	LazyValue struct {
 		mu      sync.Mutex
 		factory LazyValueFactory
@@ -66,7 +68,7 @@ func (lv *LazyValue) Mutate(ctx context.Context, mutator func(v core.Value, err 
 	mutator(lv.value, lv.err)
 }
 
-// Mutate safely mutates an underlying value only if it's ready.
+// MutateIfReady safely mutates an underlying value only if it's ready.
 func (lv *LazyValue) MutateIfReady(mutator func(v core.Value, err error)) {
 	lv.mu.Lock()
 	defer lv.mu.Unlock()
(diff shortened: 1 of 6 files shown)


pkg/drivers/common/lazy.go
```diff
@@ -63,18 +52,38 @@ func (lv *LazyValue) Read(ctx context.Context) (core.Value, error) {
...
 
 	if !lv.ready {
 		lv.load(ctx)
 	}
 
-	writer(lv.value, lv.err)
+	mutator(lv.value, lv.err)
+}
+
+// Mutate safely mutates an underlying value only if it's ready.
```
3timeslazy: Should start with `MutateIfReady ...`

kaixinbuyu:
is ferret 0.7 has the same dead lock bug?
I'm using 0.7 for some reasons.
The doc.WaitForNavigation() (0.7) func somethings raise timeout error,sometimes it runs ok.

PR #334, opened 2019-07-16
Added possibility to set custom viewport size

ziflex:
Added new settings:

```
LET p = DOCUMENT(url, {
    driver: "cdp",
    viewport: {
        width: 1024,
        height: 768,
        landscape: false,
        mobile: false,
        scaleFactor: 1
   }
})
```

### DIFF
--- a/pkg/drivers/cdp/driver.go
+++ b/pkg/drivers/cdp/driver.go
@@ -20,6 +20,11 @@ const DriverName = "cdp"
 const BlankPageURL = "about:blank"
 const DefaultTimeout = 5000 * time.Millisecond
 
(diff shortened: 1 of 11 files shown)


pkg/drivers/cdp/driver.go
```diff
@@ -20,6 +20,11 @@ const DriverName = "cdp"
 const BlankPageURL = "about:blank"
 const DefaultTimeout = 5000 * time.Millisecond
 
+var defaultScreenSize = &drivers.ScreenSize{
```
golangcibot: `defaultScreenSize` is unused (from `deadcode`)

Issue #259, opened 2019-03-14
FQL function namespaces


ziflex:
As amount of functions grows, there might be cases of function names collision.
We need to provide possibility to register these functions within different namespaces.

[This is how AQL supports it](https://docs.arangodb.com/3.4/Drivers/Java/Reference/Database/AqlUserFunctions.html).

There are several questions that need to be answered:
- What syntax? (```NAMESPACE::NAMESPACE::FUNCTION```?)
- How to register namespaces?
- Should we import namespaces in a query or it should be available automatically?

ziflex:
- Syntax is going to be ``NAMESPACE::NAMESPACE::FUNCTION``
- Registrations is the same as the registration of a regular function.
```go
err := c.RegisterFunction("T::UTILS::SPY", func(_ context.Context, _ ...core.Value) (core.Value, error) {})
```
- No imports in queries. Used as a regular pre-registered function.

PR #336, opened 2019-07-18
Added namespace builder

ziflex:
Addition to #259

### DIFF
--- a/pkg/compiler/namespace.go
+++ b/pkg/compiler/namespace.go
@@ -3,8 +3,11 @@ package compiler
 import (
 	"github.com/MontFerret/ferret/pkg/runtime/core"
 	"github.com/pkg/errors"
+	"regexp"
 	"strings"
 )
+
+var fnNameValidation = regexp.MustCompile("^[a-zA-Z]+[a-zA-Z0-9_]*(::[a-zA-Z]+[a-zA-Z0-9_]*)*$")
 
 const emptyNS = ""
 const separator = "::"
@@ -40,6 +43,10 @@ func (rs *NamespaceBuilder) RegisterFunction(name string, fun core.Function) err
 	// validation the name
 	if strings.Contains(name, separator) {
 		return errors.Errorf("invalid function name: %s", name)
+	}
+
+	if !fnNameValidation.MatchString(nsName) {
+		return errors.Errorf("invalid function or namespace name: %s", nsName)
 	}
 
 	rs.funcs[strings.ToUpper(nsName)] = fun
--- a/pkg/compiler/namespace.go
+++ b/pkg/compiler/namespace.go
@@ -25,7 +25,7 @@ func newRootNamespace() *NamespaceBuilder {
 }
 
 func newNamespace(funcs core.Functions, name string) *NamespaceBuilder {
-	return &NamespaceBuilder{funcs, name}
+	return &NamespaceBuilder{funcs, strings.ToUpper(name)}
 }
 
 func (rs *NamespaceBuilder) Namespace(name string) core.Namespace {
--- a/pkg/compiler/namespace.go
+++ b/pkg/compiler/namespace.go
@@ -1,0 +1,90 @@
+package compiler
+
+import (
+	"github.com/MontFerret/ferret/pkg/runtime/core"
+	"github.com/pkg/errors"
+	"strings"
+)
+
+const emptyNS = ""
+const separator = "::"
+
+type NamespaceBuilder struct {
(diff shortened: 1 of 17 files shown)


pkg/compiler/namespace_test.go
```diff
@@ -0,0 +1,80 @@
...
+			So(err, ShouldBeNil)
+
+			funcs := c.RegisteredFunctions()
+
+			So(len(funcs), ShouldBeGreaterThan, 1)
+		})
+
+		Convey("Namespace should return all registered functions", func() {
+			c := compiler.New()
+			err := c.Namespace("T").RegisterFunction("SPY", func(ctx context.Context, args ...core.Value) (value core.Value, e error) {
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/compiler/namespace.go
```diff
@@ -0,0 +1,97 @@
...
+func (rs *NamespaceBuilder) Namespace(name string) core.Namespace {
+	return newNamespace(rs.funcs, rs.makeFullName(name))
+}
+
+func (rs *NamespaceBuilder) RegisterFunction(name string, fun core.Function) error {
+	nsName := rs.makeFullName(name)
+	_, exists := rs.funcs[nsName]
+
+	if exists {
+		return errors.Errorf("function already exists: %s", name)
```
3timeslazy: Why not just return the nil?

pkg/stdlib/collections/lib.go
```diff
@@ -2,9 +2,9 @@ package collections
 
 import "github.com/MontFerret/ferret/pkg/runtime/core"
 
-func NewLib() map[string]core.Function {
-	return map[string]core.Function{
+func NewLib(ns core.Namespace) error {
+	return ns.RegisterFunctions(core.Functions{
```
3timeslazy: Looks weird that libraries takes namespaces and register themselves. 

In my opinion, would be better to pass libraries into namespaces.

ziflex: The idea behind this decision is that a library can create nested namespaces within a given one. Plus, in some scenarios you might want to use a specific namespace for a specific library.

pkg/compiler/namespace.go
ziflex: nil of what?

3timeslazy: of error

ziflex: why?

pkg/stdlib/collections/lib.go
3timeslazy: What about returning the new namespace?

pkg/compiler/namespace.go
3timeslazy: because function is already registered and it doesn't look like an error

pkg/compiler/namespace.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	// validation the name
+	if strings.Contains(name, separator) {
```
3timeslazy: `fnNameValidation.MatchString(nsName)` enough for validation because `fnNameValidation` match such cases as `FUNC`, `NAMESPACE::FUNC` etc. and doesn't match `FUNC::`

3timeslazy: So I think it can be removed

pkg/compiler/namespace.go
```diff
@@ -0,0 +1,97 @@
...
+	ns.funcs = make(core.Functions)
+
+	return ns
+}
+
+func newNamespace(funcs core.Functions, name string) *NamespaceBuilder {
+	return &NamespaceBuilder{funcs, strings.ToUpper(name)}
+}
+
+func (rs *NamespaceBuilder) Namespace(name string) core.Namespace {
```
3timeslazy: why shortcut for `NamespaceBuilder` is `rs` but not `nb` or `builder` for example?

pkg/compiler/namespace.go
ziflex: This is the point of the error. You get an error if you try to register a function with already used name. (Name collisions).

pkg/compiler/namespace.go
ziflex: Ah, because ``NamespaceBuilder`` was named differently in early iterations.

pkg/compiler/namespace.go
3timeslazy: Error is better. You'he right.

PR #341, opened 2019-07-25
Added errors to ClickX methods and added existence check

ziflex:
In 0.7 ``CLICK`` and ``CLICK_ALL`` was doing a check before clicking and was returning a boolean value indicating whether a target element was clicked.
In 0.8 this logic got broken and the check is not performed.

### DIFF
--- a/pkg/stdlib/html/pagination.go
+++ b/pkg/stdlib/html/pagination.go
@@ -95,11 +95,11 @@ func (i *PagingIterator) Next(ctx context.Context) (core.Value, core.Value, erro
 	exists, err := i.document.ExistsBySelector(ctx, i.selector)
 
 	if err != nil {
-		return values.ZeroInt, values.ZeroInt, err
+		return values.None, values.None, err
 	}
 
 	if !exists {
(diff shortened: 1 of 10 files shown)


pkg/stdlib/html/pagination.go
```diff
@@ -92,11 +92,17 @@ func (i *PagingIterator) Next(ctx context.Context) (core.Value, core.Value, erro
 		return values.ZeroInt, values.ZeroInt, nil
 	}
 
-	if !i.document.ExistsBySelector(ctx, i.selector) {
+	exists, err := i.document.ExistsBySelector(ctx, i.selector)
+
+	if err != nil {
+		return values.ZeroInt, values.ZeroInt, err
```
3timeslazy: Maybe return `values.None` instead `values.ZeroInt`?

ziflex: It does not really matter, since an execution gets terminated because of the returned error.

3timeslazy: To be consistent I'm suggest to return `values.None` every time error occured.

PR #346, opened 2019-08-02
Bugfix/logging

ziflex:


pkg/runtime/logging/logger.go
```diff
@@ -27,19 +28,17 @@ const (
...
 )
 
-func WithContext(ctx context.Context, opts *Options) context.Context {
-	id, err := uuid.NewV4()
+func WithContext(ctx context.Context, opts Options) context.Context {
+	c := zerolog.New(opts.Writer).With().Timestamp()
 
-	if err != nil {
-		panic(err)
+	if len(opts.Fields) > 0 {
```
3timeslazy: Think we can iterate through fields without check

Issue #362, opened 2019-08-25
Clear input data


ziflex:
At this moment ``INPUT`` appends text to a value that is already in a target input field and there is not way to reset its value.

We need a special function for this.
Something like ``INPUT_CLEAR`` or something like this.

PR #366, opened 2019-08-30
Added INPUT_CLEAR function

ziflex:


pkg/drivers/cdp/input/keyboard.go
```diff
@@ -33,14 +63,14 @@ func (k *Keyboard) Up(ctx context.Context, char string) error {
 }
 
 func (k *Keyboard) Type(ctx context.Context, text string, delay int) error {
-	for _, ch := range text {
+	for _, ch := range []rune(text) {
```
golangcibot: S1029: should range over string, not []rune(string) (from `gosimple`)

ziflex:
closes #362

PR #368, opened 2019-08-30
Fixed access to a member property right after a function call

### LINKED ISSUE #335: Unable to access to a member property after a function call
**Describe the bug**
Unable to access to a member property after a function call

**To Reproduce**
Here is a query:
```
LET items = [{name: "foo"}]
RETURN FIRST(items).name
```

**Expected behavior**
Must return "foo" string.

**Actual behavior**
Returns an entire object.

ziflex:
Closes #335


pkg/compiler/visitor.go
```diff
@@ -736,15 +736,39 @@ func (v *visitor) doVisitForExpressionStatement(ctx *fql.ForExpressionStatementC
...
+	var source core.Expression
+	var children []antlr.Tree
 
-	_, err := scope.GetVariable(varName)
+	identifier := ctx.Identifier()
 
-	if err != nil {
-		return nil, err
+	if identifier != nil {
+		varName := ctx.Identifier().GetText()
```
3timeslazy: Why not just `identifier.GetText()`?

ziflex: 🤷‍♂️for clarity, I guess 😄

ziflex: Because I use the variable to do an existence check

PR #372, opened 2019-09-04
Feature/#360 default driver params

### LINKED ISSUE #360: Default headers and cookies in drivers
There are some scenarios, when we need to execute queries using some predefined headers or/and cookies.
In order to avoid tedious copy/pasting, it would be really great to create a copy of a driver with these predefined values.

If a script provides a value for one of the default headers/cookies, it should override the default one.

ziflex:
Closes #360


pkg/drivers/http/page.go
```diff
@@ -2,24 +2,27 @@ package http
...
 	"github.com/MontFerret/ferret/pkg/runtime/values"
-	"github.com/PuerkitoBio/goquery"
-	"hash/fnv"
 )
 
 type HTMLPage struct {
 	document *HTMLDocument
-	cookies  []drivers.HTTPCookie
+	cookies  drivers.HTTPCookies
+	headers  drivers.HTTPHeaders
```
golangcibot: `headers` is unused (from `structcheck`)

PR #386, opened 2019-09-11
update GoReleaser config file

clock21am:


ziflex:
@clock21am hey, thank you for your PR!
It looks good, but I have one question: [in the example](https://goreleaser.com/deprecations/#archive) they use ``id`` property, but your PR does not contain it. Will it work without it?

clock21am:
have verified from the https://github.com/goreleaser/goreleaser/blob/master/.goreleaser.yml

ziflex:
Awesome! Thank you for your help!

PR #388, opened 2019-09-15
Load HTML pages from memory 

clock21am:
Here is the changes,  I have made. I tried it. but still unsure about the conclusion about the object of this bug. please take a look.
And I had a question how are you debugging the code. I am using VScode but facing a bit issue over there. I have macOS also.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
            "type": "go",
            "request": "launch",
            "mode": "remote",
-           "remotePath": "/Users/rajdeepkaur/code/ferret",
+           "remotePath": "/Users/..User_Name/code/ferret",
            "port": 9222,
            "host": "0.0.0.0",
            "program": "${workspaceRoot}",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+    "version": "0.2.0",
+    "confi'lgurations": [
+       {
+           "name": "Remote Docker",
+           "type": "go",
+           "request": "launch",
+           "mode": "remote",
+           "remotePath": "/Users/rajdeepkaur/code/ferret",
+           "port": 9222,
+           "host": "0.0.0.0",
+           "program": "${workspaceRoot}",
+           "env": {},
+           "args": []
+       }
+    ]
+}
--- a/pkg/stdlib/html/html_parse.go
+++ b/pkg/stdlib/html/html_parse.go
@@ -11,12 +11,9 @@ import (
 	"github.com/pkg/errors"
 )
 
-func HtmlParse(ctx context.Context, args ...core.Value) (core.Value, error) {
+func HTMLParse(ctx context.Context, args ...core.Value) (core.Value, error) {
 	err := core.ValidateType(args[0], types.String)
 
-	if err != nil {
-		return values.False, err
-	}
 	str := values.ToString(args[1])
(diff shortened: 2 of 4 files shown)


pkg/stdlib/html/html_parse.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/MontFerret/ferret/pkg/runtime/values/types"
+	"github.com/PuerkitoBio/goquery"
+	"github.com/pkg/errors"
+)
+
+func HtmlParse(ctx context.Context, args ...core.Value) (core.Value, error) {
+	err := core.ValidateType(args[0], types.String)
+
+	if err != nil {
+		return values.False, err
```
3timeslazy: Why are you returning `return values.False, err` here and `return nil, errors.Wrap(err, "failed to parse a document")` below? Doesn't look consistent.

pkg/stdlib/html/html_parse.go
```diff
@@ -0,0 +1,30 @@
...
+	"context"
+
+	"github.com/MontFerret/ferret/pkg/runtime/core"
+	"github.com/MontFerret/ferret/pkg/runtime/values"
+	"github.com/MontFerret/ferret/pkg/runtime/values/types"
+	"github.com/PuerkitoBio/goquery"
+	"github.com/pkg/errors"
+)
+
+func HtmlParse(ctx context.Context, args ...core.Value) (core.Value, error) {
```
3timeslazy: Better rename to `HTMLParse`

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
+    "version": "0.2.0",
+    "confi'lgurations": [
+       {
+           "name": "Remote Docker",
+           "type": "go",
+           "request": "launch",
+           "mode": "remote",
+           "remotePath": "/Users/rajdeepkaur/code/ferret",
```
3timeslazy: These are some local settings, right?

pkg/stdlib/html/html_parse.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/MontFerret/ferret/pkg/runtime/core"
+	"github.com/MontFerret/ferret/pkg/runtime/values"
+	"github.com/MontFerret/ferret/pkg/runtime/values/types"
+	"github.com/PuerkitoBio/goquery"
+	"github.com/pkg/errors"
+)
+
+func HTMLParse(ctx context.Context, args ...core.Value) (core.Value, error) {
+	err := core.ValidateType(args[0], types.String)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

ziflex:
Sorry for being silent for so long - I was on vacation :)

Ok, I see what you have here and I think the best way would be doing it in several steps.
- First of all, we need to extends the [``Driver``](https://github.com/MontFerret/ferret/blob/master/pkg/drivers/driver.go#L17) interface by adding a new ``.Parse(ctx context.Context, content []byte) (HTMLPage, error)`` method.
- Once you update the driver, change the function you created by using the interface, not a particular implementation - standard library works only with interfaces. You can take a look how to get a driver [here](https://github.com/MontFerret/ferret/blob/master/pkg/stdlib/html/document.go#L65). Also, you will need to allow users specify what driver to use by passing a param object.
- After that, implement document parsing in each driver. The current implementation, that this PR has will go to ``http`` driver.

PR #390, opened 2019-09-22
Bugfixes in http driver functions that utilize goquery

gyy52380:
There were some bugs in the http driver's function that used goquery (`ELEMENT_EXISTS`, `INNER_TEXT`):
- goquery's `selection.Find()` always returns a non-null selection object even when empty. `selection.Length()` must be used to check for contents.
- goquery's `selection.Closest()` looks at selection's parents instead of children, which is the expected behaviour for `ELEMENT_EXISTS(element, selector)`

This bugfix might break a lot of people's code; for example, `INNER_TEXT` often returned an empty string when it should've returned an error instead.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,6 +1,8 @@
 github.com/Masterminds/glide v0.13.2/go.mod h1:STyF5vcenH/rUqTEv+/hBXlSTo7KYwg2oc2f4tzPWic=
 github.com/Masterminds/semver v1.4.2/go.mod h1:MB6lktGJrhw8PrUyiEoblNEGEQ+RzHPF078ddwwvV3Y=
 github.com/Masterminds/vcs v1.13.0/go.mod h1:N09YCmOQr6RLxC6UNHzuVwAdodYbbnycGHSmwVJjcKA=
+github.com/MontFerret/ferret v0.9.0 h1:eRnwpp2VIh+3/u6x7kAG0tR8mr3/PBnkqM8Ag9uocZ8=
+github.com/MontFerret/ferret v0.9.0/go.mod h1:OBon0OYEs0sdhnuTi+M4UB7ya4ckBry4Wx/++EVSJ3M=
 github.com/PuerkitoBio/goquery v1.5.0 h1:uGvmFXOA73IKluu/F84Xd1tt/z07GYm8X49XKHP7EJk=
 github.com/PuerkitoBio/goquery v1.5.0/go.mod h1:qD2PgZ9lccMbQlc7eEOjaeRlFQON7xY8kdmcsrnKqMg=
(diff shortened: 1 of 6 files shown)


ziflex:
Looks good! Just need to fix a failing e2e test.

Regarding breaking changes - we haven't gotten to 1.0, so we can break the things :)

gyy52380:
[This](https://github.com/MontFerret/ferret/blob/master/e2e/tests/el_element_exists_d.fql) test fails. 

As far as I can see, this test should be performed dynamically by using `DOCUMENT(url, true)` instead of `DOCUMENT(url)` which is currently in the test. When the document isn't dynamic, both conditions should be false. 

The test used to find because `ELEMENT_EXISTS()` for http driver always returned true.

go.sum
```diff
@@ -1,6 +1,8 @@
 github.com/Masterminds/glide v0.13.2/go.mod h1:STyF5vcenH/rUqTEv+/hBXlSTo7KYwg2oc2f4tzPWic=
 github.com/Masterminds/semver v1.4.2/go.mod h1:MB6lktGJrhw8PrUyiEoblNEGEQ+RzHPF078ddwwvV3Y=
 github.com/Masterminds/vcs v1.13.0/go.mod h1:N09YCmOQr6RLxC6UNHzuVwAdodYbbnycGHSmwVJjcKA=
+github.com/MontFerret/ferret v0.9.0 h1:eRnwpp2VIh+3/u6x7kAG0tR8mr3/PBnkqM8Ag9uocZ8=
```
ziflex: I guess this came from your fork?

pkg/drivers/http/driver.go
```diff
@@ -152,7 +152,7 @@ func (drv *Driver) Open(ctx context.Context, params drivers.Params) (drivers.HTM
 
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusOK {
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusInternalServerError {
```
ziflex: Why did you add this extra condition?

go.sum
gyy52380: Yes, sorry, I removed this.

pkg/drivers/http/driver.go
gyy52380: This also came from my local fork, because of some testing I was doing. I removed it.

Side note: an option to add custom "retry codes" (like what scrapy offers) would be good. Some sites return perfectly valid html but respond with status 500. I usually want to consider a request as failed only if it returned 200 or 202.

ziflex: Yeah, sure, you can def think about it.

ziflex:
Thank you for your time and effort!

Issue #382, opened 2019-09-07
Add response information to HTML page


ziflex:
Add a response data to an HTML page that contains information about response status, headers and etc.

- [x] HTTP driver
- [ ] CDP driver

PR #391, opened 2019-10-03
feature: issue_382

3timeslazy:


pkg/drivers/response.go
```diff
@@ -0,0 +1,103 @@
...
+	return *(&resp)
+}
+
+func (resp *HTTPResponse) Hash() uint64 {
+	return values.Parse(resp).Hash()
+}
+
+// responseMarshal is a structure that repeats HTTPResponse. It allows
+// easily Marshal the HTTPResponse object.
+type responseMarshal struct {
```
ziflex: Why did you decide to go with this approach? Did you consider other options?

For example, using map[string]interface{} for marshalling in ``HTTPResponse`` struct?

```go
func (resp *HTTPResponse) MarshalJSON() ([]byte, error) {
    j := map[string]interface{}{
        "status_code": resp.StatusCode,
        "status": resp.Status,
        "headers": resp.Headers
    }
    
    return json.Marshal(j)
}
```

Or, maybe embedding would work?

```go
type HTTPResponseBase struct {
	StatusCode int         `json:"status_code"`
	Status     string      `json:"status"`
	Headers    HTTPHeaders `json:"headers"`
}

type HTTPResponse struct {
	HTTPResponseBase
}

func (resp *HTTPResponse) MarshalJSON() ([]byte, error) {
	if resp == nil {
		return json.Marshal(values.None)
	}

	return json.Marshal(resp.HTTPResponseBase)
}

```

go.sum
```diff
@@ -64,6 +64,8 @@ github.com/valyala/bytebufferpool v1.0.0 h1:GqA5TC/0021Y/b9FG4Oi9Mr3q7XYx6Kllzaw
 github.com/valyala/bytebufferpool v1.0.0/go.mod h1:6bBcMArwyJ5K/AmCkWv1jt77kVWyCJ6HpOuEn7z0Csc=
 github.com/valyala/fasttemplate v1.0.1 h1:tY9CJiPnMXf1ERmG2EyK7gNUd+c6RKGD0IfU8WdUSz8=
 github.com/valyala/fasttemplate v1.0.1/go.mod h1:UQGH1tvbgY+Nz5t2n7tXsz52dQxojPUpymEIMZ47gx8=
+github.com/yudai/pp v2.0.1+incompatible h1:Q4//iY4pNF6yPLZIigmvcl7k/bPgrcTPIFIcmawg5bI=
+github.com/yudai/pp v2.0.1+incompatible/go.mod h1:PuxR/8QJ7cyCkFp/aUDS+JY727OFEZkTdatxwunjIkc=
```
ziflex: What is this for?

pkg/drivers/response.go
3timeslazy: I decided to use this approach because in my opinion this is best way to keep HTTPResponse and it's marshal form consistent.

1. Why not using map. Because if another field appears in the HTTPResponse, it’s easy to forget to add it to the map.

2. Why not embedding. Embedded struct this is the same with current approach, but have a single difference.
```go
// Init using different struct
r := drivers.HTTPResponse{
    StatusCode: resp.StatusCode,
    Status: resp.Status,
    Headers: drivers.HTTPHeaders(resp.Header),
}

// Init using embedded struct - got one more nested level
r := drivers.HTTPResponse{
    HTTPResponseBase: drivers.HTTPResponseBase{
        StatusCode: resp.StatusCode,
        Status: resp.Status,
        Headers: drivers.HTTPHeaders(resp.Header),
    }
}
```

3timeslazy: Code will not compile if `HTTPResponse` and `responseMarshal` would have different fields. 
```go
// there will be a compile error
return json.Marshal(responseMarshal(*resp))
```

go.sum
3timeslazy: Some debug dependency

ziflex: Do we need it?

3timeslazy: No. I removed it.

pkg/drivers/response.go
ziflex: > I decided to use this approach because in my opinion this is best way to keep HTTPResponse and it's marshal form consistent.
> 
> 1. Why not using map. Because if another field appears in the HTTPResponse, it’s easy to forget to add it to the map.
> 2. Why not embedding. Embedded struct this is the same with current approach, but have a single difference.
> 
> ```go
> // Init using different struct
> r := drivers.HTTPResponse{
>     StatusCode: resp.StatusCode,
>     Status: resp.Status,
>     Headers: drivers.HTTPHeaders(resp.Header),
> }
> 
> // Init using embedded struct - got one more nested level
> r := drivers.HTTPResponse{
>     HTTPResponseBase: drivers.HTTPResponseBase{
>         StatusCode: resp.StatusCode,
>         Status: resp.Status,
>         Headers: drivers.HTTPHeaders(resp.Header),
>     }
> }
> ```

I do not see a problem in a nested struct - we have just 2 places where it will be used and it will unlikely change. But the benefit is that you don't have to have 2 structs with identical fields.

3timeslazy: `HTTPResponseBase` implies that somewhere there must be `HTTPResponse` or `HTTPResponseExtended` or something like that. 

This approach provides more clear and small package API. Keeping both `HTTPResponseBase` and `HTTPResponse` can be confusing.

pkg/drivers/response.go
```diff
@@ -0,0 +1,100 @@
...
+func (resp *HTTPResponse) String() string {
+	return resp.Status
+}
+
+func (resp *HTTPResponse) Compare(other core.Value) int64 {
+	if other.Type() != HTTPResponseType {
+		return Compare(HTTPResponseType, other.Type())
+	}
+
+	// it's a safe cast because HTTPResponse doesn't implement
```
ziflex: does NOT?

3timeslazy: But *HTTPResponse implements

3timeslazy: add more descriptive comment https://github.com/MontFerret/ferret/pull/391/commits/338bcc5cd43e2554c1189263cb910abd9db68f8e

ziflex:
Great job!  👍

ziflex:
#382

PR #393, opened 2019-10-09
Make FQL keywords case-insensitive

3timeslazy:


pkg/parser/parser.go
```diff
@@ -12,7 +13,11 @@ type Parser struct {
 
 func New(query string) *Parser {
 	input := antlr.NewInputStream(query)
-	lexer := fql.NewFqlLexer(input)
+	// converts tokens to upper case, so now it doesn’t matter
```
ziflex: To upper case?

3timeslazy: Yes. You can type key-words in any case because they all will be converted to upper case before parsing.

ziflex:
Does this PR supports namespaces and functions on both cases?

3timeslazy:
> Does this PR supports namespaces and functions on both cases?

pkg/stdlib/datetime/lib.go
```diff
@@ -2,8 +2,8 @@ package datetime
 
 import "github.com/MontFerret/ferret/pkg/runtime/core"
 
-func NewLib() map[string]core.Function {
-	return map[string]core.Function{
+func NewLib() core.FunctionsMap {
```
3timeslazy: Why `RegisterLib` function doesn't register `datetime` fucntions? https://github.com/MontFerret/ferret/blob/e96012140d89b9aa09f3dc15ad34ba24a08a8aff/pkg/stdlib/lib.go#L15-L45

Looks like a bug

ziflex:
Great job!

Issue #385, opened 2019-09-10
Load HTML pages from memory


ziflex:
Ferret used to have this feature but during one of the refactorings it was decided to remove it.

Let's bring it back.

- Add ``Parse(ctx context.Context, content []byte) (HTMLPage, error)`` method to ``Driver`` interface.

- Use [this method](https://chromedevtools.github.io/devtools-protocol/tot/Page#method-setDocumentContent) in CDP driver.

- Use ``goquery.NewDocumentFromReader`` in HTTP driver.

- Add ``HTML_PARSE`` function to ``html`` library.

clock21am:
hey @ziflex  could you please explain this issue a bit more in detail?

ziflex:
@clock21am sure!

In Ferret, we have a concept of HTTP drivers that represent different implementations of access to HTML pages.
[Here](https://github.com/MontFerret/ferret/blob/master/pkg/drivers/driver.go) is an interface: 
```go
type Driver interface {
    io.Closer
    Name() string
    Open(ctx context.Context, params Params) (HTMLPage, error)
}
```

Currently, we have 2 drivers:

- Static driver, that retrieves pages by a regular HTTP call and stores them in memory.
- Dynamic driver, that uses CDP (Chrome Devtools Protocol) to control Chrome/Chromium remotely and provides an access to DOM through it.

And for now, there is only one way how to load a page - you should call a hosting server i.e. make a request. 
In case of a static driver, we just make an HTTP GET request to a given url. 
In case of dynamic one, we open a tab in a browser and navigate it to a target url.

All that happens when you call ``driver.Open`` method.
In both scenarios, the webpage must be hosted.

What this task for is to add an ability to load a pre-fetched HTML file into a one of the above drivers.

The idea of how it should be implemented is in the description.

clock21am:
Hey @ziflex thanks alot for the deep description, Let me take a look.

laylwintthu396:
I Love Children.

patrikhl:
Any progress on this? Otherwise I will take a look at it.

clock21am:
Hey so sorry, was occupied with the office stuffs the last week, will take a look on it this weekend. if you fix it before go ahead.

PR #394, opened 2019-10-09
 Load HTML pages from memory #385 

patrikhl:
For https://github.com/MontFerret/ferret/issues/385

The http version is tested and works with examples/static-page.fql.

The cdp version has the problem that the default write buffer size is 4096 so SetDocumentContent cannot load files larger than that (it returns the error: "cdp.Page: SetDocumentContent: cdp.Target: SendMessageToTarget: rpcc: message too large...").

This can be fixed by setting the buffer size in Driver.init(...), but I have not looked into it further.

The cdp version has been tested with the following simple example:
```
<!DOCTYPE html>
<html>
<body>Test</body>
</html>
```
```
LET doc = HTML_PARSE("test.html", {driver:"cdp"})
LET str = ELEMENT(doc, "body")
RETURN str
```

### DIFF
--- a/pkg/drivers/cdp/driver.go
+++ b/pkg/drivers/cdp/driver.go
@@ -2,15 +2,20 @@ package cdp
 
 import (
 	"context"
+	"sync"
+
 	"github.com/mafredri/cdp"
 	"github.com/mafredri/cdp/devtool"
+	"github.com/mafredri/cdp/protocol/network"
+	"github.com/mafredri/cdp/protocol/page"
 	"github.com/mafredri/cdp/protocol/target"
 	"github.com/mafredri/cdp/rpcc"
 	"github.com/mafredri/cdp/session"
 	"github.com/pkg/errors"
-	"sync"
 
 	"github.com/MontFerret/ferret/pkg/drivers"
+	"github.com/MontFerret/ferret/pkg/drivers/cdp/events"
+	"github.com/MontFerret/ferret/pkg/drivers/cdp/input"
 	"github.com/MontFerret/ferret/pkg/runtime/logging"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/stdlib/html/html_parse.go
```diff
@@ -0,0 +1,59 @@
...
+		p, err := newPageLoadParams("about:blank", args[1])
+
+		if err != nil {
+			return values.None, err
+		}
+
+		params = p
+	}
+
+	b, err := ioutil.ReadFile(args[0].String())
```
ziflex: I would prefer to pass in content. I added an issue for IO function for such cases.

pkg/drivers/cdp/driver.go
```diff
@@ -135,6 +140,131 @@ func (drv *Driver) Open(ctx context.Context, params drivers.Params) (drivers.HTM
 	return LoadHTMLPage(ctx, conn, params)
 }
 
+func (drv *Driver) Parse(ctx context.Context, content []byte) (drivers.HTMLPage, error) {
```
ziflex: It feels that the code is copied from the ``Open`` function :)

Can we refactor the driver to avoid copying this?

pkg/drivers/cdp/driver.go
```diff
@@ -135,6 +140,131 @@ func (drv *Driver) Open(ctx context.Context, params drivers.Params) (drivers.HTM
...
+			return client.Page.SetBypassCSP(ctx, page.NewSetBypassCSPArgs(true))
+		},
+	)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// Get frame tree
+	ft, err := client.Page.GetFrameTree(ctx)
```
ziflex: I guess we need to invoke this method after setting a document's content.

ziflex:
Hey, thanks for the PR! 
There are some things that I would like to change, but other than that it looks good.

Could you try to increase the buffer size?

pkg/drivers/cdp/driver.go
patrikhl: Most of the code is a boiled down version of LoadHTMLPage(...). But yes, I can move the code that is shared between Open(...) and Parse(...) to a new function and call that from both of them.

pkg/drivers/cdp/driver.go
patrikhl: SetDocumentContent(...) takes the frameID as an argument, so it is called before to get that.

pkg/stdlib/html/html_parse.go
patrikhl: Should I change it, or leave it as it is until the IO functions get implemented?

patrikhl:
Hi, I will look into the changes tomorrow.

Changing the buffer size is just a case of passing the size to rpcc.DialContext(...) in Driver.init(...), so an option would be to change init(...) to take a list of arguments that can be passed to DialContext(...):

```func (drv *Driver) init(ctx context.Context, dialOpts ...rpcc.DialOption) error```

pkg/stdlib/html/html_parse.go
ziflex: I would change it. Pass content as a binary value.

When #395 gets closed - you will be able to read a file from disk and pass output to the function.

pkg/drivers/cdp/driver.go
ziflex: Ok, thanks for explanation!

3timeslazy: Write a comment with explanation from this thread, please

pkg/stdlib/html/html_parse.go
3timeslazy: @ziflex @patrikhl you can [read](https://github.com/MontFerret/ferret/pull/403) from disk disk now

ziflex:
Hey @patrikhl, any progress on this?

PR #396, opened 2019-10-11
Added params check before execution

### LINKED ISSUE #373: Check param values before execution
Make sure that all values are provided for all existing params in a script before its execution.

ziflex:
Closes #373

### DIFF
--- a/pkg/compiler/scope.go
+++ b/pkg/compiler/scope.go
@@ -2,47 +2,56 @@ package compiler
 
 import (
 	"github.com/MontFerret/ferret/pkg/runtime/core"
-	"github.com/MontFerret/ferret/pkg/runtime/values/types"
 )
 
 type (
+	globalScope struct {
+		params map[string]struct{}
+	}
+
 	scope struct {
+		global *globalScope
 		parent *scope
-		vars   map[string]core.Type
+		vars   map[string]struct{}
 	}
 )
 
-func newRootScope() *scope {
+func newGlobalScope() *globalScope {
+	return &globalScope{
+		params: map[string]struct{}{},
+	}
+}
+
+func newRootScope(global *globalScope) *scope {
 	return &scope{
-		vars: make(map[string]core.Type),
+		global: global,
+		vars:   make(map[string]struct{}),
 	}
 }
 
 func newScope(parent *scope) *scope {
-	s := newRootScope()
+	s := newRootScope(parent.global)
 	s.parent = parent
 
 	return s
 }
 
-func (s *scope) GetVariable(name string) (core.Type, error) {
-	local, exists := s.vars[name]
+func (s *scope) AddParam(name string) {
+	s.global.params[name] = struct{}{}
+}
+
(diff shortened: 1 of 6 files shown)


pkg/compiler/scope.go
```diff
@@ -53,7 +62,7 @@ func (s *scope) SetVariable(name string) error {
 	}
 
 	// TODO: add type detection
```
3timeslazy: There is no types now

ziflex: Todo for the future. Until we implement it, no reason to use type as a map value.

pkg/runtime/program.go
```diff
@@ -24,16 +26,33 @@ func NewProgram(src string, body core.Expression) (*Program, error) {
...
 }
 
 func (p *Program) Source() string {
 	return p.src
 }
 
+func (p *Program) Params() []string {
+	res := make([]string, 0, len(p.params))
+
+	for name := range p.params {
```
3timeslazy: You can use `copy` instead
```golang
res := make([]string, len(p.params))
copy(res, p.params)
return res
```

pkg/runtime/program.go
```diff
@@ -92,3 +111,29 @@ func (p *Program) MustRun(ctx context.Context, setters ...Option) []byte {
...
+		return nil
+	}
+
+	var missedParams []string
+
+	for n := range p.params {
+		_, exists := opts.params[n]
+
+		if !exists {
+			if missedParams == nil {
```
3timeslazy: There are no reasons to check `missedParams == nil` because `append(nil-slice, elem)` returns `slice{elem}` and doesn't panic

ziflex: Good to know, thanks!

pkg/runtime/program.go
ziflex: Always forget about ``copy`` function!

ziflex: Actually, it does not work like this. ``params`` is a map.

pkg/runtime/program.go
ziflex: But, won't it re-allocate a new slice on every append?

pkg/runtime/program.go
3timeslazy: Sorry, my bad

pkg/runtime/program.go
3timeslazy: It will but not on every append. Finally, there is nothing prevents you from doing 
```go
missedParams :=  make([]string, 0, len(p.params))

for n := range p.params { ...
```

ziflex: That's what I'm doing, but lazily. Why do I need to allocate it if there are might be no errors?
There is a chance that escape analysis will optimize the code though...

3timeslazy: If you doing this because there are might be no errors, it's OK but need a comment. Otherwise it looks like an unnecessary check.

ziflex: Added a comment

PR #398, opened 2019-10-13
allowed http response codes option

gyy52380:
Very often pages will return valid html even if the http response is 500 - internal server error. Even if the html is not what the user might have expected, some html content will exist. I added this change to my local fork because of how common it is to see a "valid" response code 500. 

Feel free to discard this change if it seems hacky. It might be better to add some `WithAllowedHttpResponseCodes` option to the http driver initialisation, but I didn't find that necessary.


gyy52380:
I decided to add the option to the driver after all. The previous solution was too hacky.

ziflex:
Looks good to me.
@3timeslazy any comments?

pkg/drivers/http/options.go
```diff
@@ -125,3 +129,21 @@ func WithCookies(cookies []drivers.HTTPCookie) Option {
...
+		if opts.AllowedHTTPCodes == nil {
+			opts.AllowedHTTPCodes = make([]int, 0, 1)
+		}
+		opts.AllowedHTTPCodes = append(opts.AllowedHTTPCodes, httpCode)
+	}
+}
+
+func WithAllowedHTTPCodes(httpCodes []int) Option {
+	return func(opts *Options) {
+		if opts.AllowedHTTPCodes == nil {
```
3timeslazy: It’s not necessary to check slice is nil or not before append

pkg/drivers/http/options.go
```diff
@@ -125,3 +129,21 @@ func WithCookies(cookies []drivers.HTTPCookie) Option {
 		}
 	}
 }
+
+func WithAllowedHTTPCode(httpCode int) Option {
```
3timeslazy: I think WithAllowedHTTPCode is excess here. We already have WithAllowedHTTPCodes

pkg/drivers/http/driver.go
```diff
@@ -152,7 +152,15 @@ func (drv *Driver) Open(ctx context.Context, params drivers.Params) (drivers.HTM
 
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusOK {
```
3timeslazy: I think we should add a small function that detects allowed codes

pkg/drivers/http/options.go
gyy52380: Thanks, fixed.

pkg/drivers/http/options.go
gyy52380: I added two functions because all the other Options come in two variants, one singular and one plural. But yes, in my opinion one vararg function would be cleaner: `func WithAllowedHTTPCodes(httpCodes ...int)`. I left it as-is for now.

pkg/drivers/http/driver.go
gyy52380: I placed the check into an unexported function.

3timeslazy:
Just one more thing. Can you remove this comment, pls? Thanks to this PR, it is not relevant
https://github.com/MontFerret/ferret/blob/b9a55a7582d1e42e1e7ffecadc0e1ac3af103a62/pkg/drivers/http/driver.go#L171

pkg/drivers/http/options.go
ziflex: I think 2 functions is ok.

gyy52380:
I removed the comment.

gyy52380:
Uh, so GolangCI wont run for some reason. It says that function main was redeclared in the example/ folder. Can someone help with this? I have no idea why this suddenly doesnt pass, even though it passes travis just fine.

pkg/drivers/http/options.go
```diff
@@ -26,6 +29,7 @@ func newOptions(setters []Option) *Options {
 	opts.Backoff = pester.ExponentialBackoff
 	opts.Concurrency = 3
 	opts.MaxRetries = 5
+	opts.AllowedHTTPCodes = []int{stdhttp.StatusOK}
```
ziflex: Maybe ``map[int]struct{}`` instead of the slice?

You could change the existence check later like this:
```go
_, exists := drv.options.AllowedHTTPCodes[resp.StatusCode]

return true
```

gyy52380: Sure, I changed it. 

Why do you think this way is better? I think that using a map is even a bit slower because the map must be initialised and more time will be used on the hash table lookup of int than just iterating through a slice with < 20 elements in it.

pkg/drivers/http/driver.go
```diff
@@ -177,15 +177,6 @@ func (drv *Driver) Open(ctx context.Context, params drivers.Params) (drivers.HTM
 	return NewHTMLPage(doc, params.URL, &r, cookies)
 }
 
-func (drv *Driver) responseCodeAllowed(resp *http.Response) bool {
```
ziflex: No need to remove the function. It has a good descriptive name :)

ziflex:
We are almost there! Please, update formatting!

ziflex:
Great! Thank you for your time, effort and patience!

PR #402, opened 2019-10-18
Huge CDP event broker performance improvement

gyy52380:
**Changes**
Merged multiple select statements in `runLoop()` into a single one which greatly reduced mutex contention.

Changed context cancelation handling in `runLoop()` select statement to work with cdp `Ready()` channels.

**Rationale**
During my own work with the cdp package i struggled with performant simulatneous event checking and context cancelation checking.

cdp Ready() channels are closed when a message is ready, or when the event client is canceled.
If the event client is canceled, eventClient.RecvMsg() will return a non-nil error whose cause will be context.Canceled or context.DeadlineExceeded. By checking for this cause we know if the context is valid or not, and therefore whether the loop should quit. I've been using this method in my own codebase. [CDP](https://github.com/mafredri/cdp/blob/master/rpcc/conn_test.go#L201) also uses this method to check for context cancelation.

This yields performance gains that scale with the number of chrome pages. My cpu usage went from 60% for 3 pages to <1%.


gyy52380:
During my own work with the `cdp` package i struggled with performant simulatneous event checking and context cancelation checking.

cdp `Ready()` channels are closed when a message is ready, or when the event client is canceled.
If the event client is canceled, `eventClient.RecvMsg()` will return a non-nil error whose cause will be `context.Canceled` or `context.DeadlineExceeded`. By checking for this cause we know if the context is valid or not, and therefore whether the loop should quit. I've been using this method in my own codebase. [CDP](https://github.com/mafredri/cdp/blob/master/rpcc/conn_test.go#L201) uses this method to check for context cancelation.

This yields performance gains that scale with the number of chrome pages. My cpu usage went from 60% for 3 pages to <1%.

ziflex:
I like it! Yeah, it seems ``select`` has a pretty decent performance penalty :|

But, it seems that e2e tests started failing.

gyy52380:
Yes, I must be doing something wrong. I will try to fix it.

gyy52380:
I fixed it. I was checking if the wrong context was canceled.

gyy52380:
Side note: I think that your ChildNoteInserted event isn't working correctly. 

The event broker listens for `ChildNoteInserted` event. This event is raised only for the top-most appended child node. For example, if this `ul` node were appended, only one event would be raised whose nodeID would correspond to the `ul` element:

```
<ul>
    <li>1</li>
    <li>2</li>
    <li>3</li>
</ul>
```

You need to listen to the [`SetChildNode`](https://godoc.org/github.com/mafredri/cdp/protocol/dom#SetChildNodesClient) event if you want to also get notified when nested children (like `<li>` here) get appended.

Some extra code is needed to request the nested children, in my code I use something similar to this:

```
case <-ctx.eventNodeInserted.Ready():
	event, err := eventNodeInserted.Recv()
	if err != nil {
	        eventError = err
		break
	}
	ctx.nodeChan <- event.Node.NodeID

	// event contains only topmost inserted node. query all of its children.
	err = client.DOM.RequestChildNodes(timeoutCtx,dom.NewRequestChildNodesArgs(event.Node.NodeID).SetDepth(-1))
	if err != nil {
		eventError = err
		break
	}
```

Should I start a new pull request for this?

ziflex:
> Side note: I think that your ChildNoteInserted event isn't working correctly.
> 
> The event broker listens for `ChildNoteInserted` event. This event is raised only for the top-most appended child node. For example, if this `ul` node were appended, only one event would be raised whose nodeID would correspond to the `ul` element:
> 
> ```
> <ul>
>     <li>1</li>
>     <li>2</li>
>     <li>3</li>
> </ul>
> ```
> 
> You need to listen to the [`SetChildNode`](https://godoc.org/github.com/mafredri/cdp/protocol/dom#SetChildNodesClient) event if you want to also get notified when nested children (like `<li>` here) get appended.
> 
> Some extra code is needed to request the nested children, in my code I use something similar to this:
> 
> ```
> case <-ctx.eventNodeInserted.Ready():
> 	event, err := eventNodeInserted.Recv()
> 	if err != nil {
> 	        eventError = err
> 		break
> 	}
> 	ctx.nodeChan <- event.Node.NodeID
> 
> 	// event contains only topmost inserted node. query all of its children.
> 	err = client.DOM.RequestChildNodes(timeoutCtx,dom.NewRequestChildNodesArgs(event.Node.NodeID).SetDepth(-1))
> 	if err != nil {
> 		eventError = err
> 		break
> 	}
> ```
> 
> Should I start a new pull request for this?

Oh, thanks that you noticed it. Sure, any help is welcomed!

ziflex:
Thanks for your help!

Issue #395, opened 2019-10-09
Add IO functions


ziflex:
Let's add functions that implement IO operations.

We need to cover 2 areas for now:

- File system
- Network (HTTP)

Each of the areas must be in a dedicated namespace:

- ``IO::FS::READ(path: string)``, ``IO::FS::WRITE(path: string, data: []bytes)``
- ``IO::NET::HTTP::GET(url: string, params?: object)``, ``IO::NET::HTTP::POST(params: object)``, ``IO::NET::HTTP::PUT(params: object)``, ``IO::NET::HTTP::DELETE(url: string, params?: object)``

Did I miss something?

More detailed API will be discussed later.

Here is a list of things we need to add for starters:
- [x] IO::FS::READ
- [x] IO::FS::WRITE
- [ ] IO::FS::DELETE (do we need it?)
- [ ] IO::NET::HTTP::GET
- [ ] IO::NET::HTTP::POST
- [ ] IO::NET::HTTP::PUT
- [ ] IO::NET::HTTP::DELETE

3timeslazy:
Can take this

ziflex:
@3timeslazy any update on this?

3timeslazy:
@ziflex in progress

PR #405, opened 2019-10-21
add IO::FS:WRITE function

3timeslazy:
Updates #395

### DIFF
--- a/pkg/stdlib/io/fs/write.go
+++ b/pkg/stdlib/io/fs/write.go
@@ -1,0 +1,154 @@
+package fs
+
+import (
+	"context"
+	"os"
+	"sort"
+
+	"github.com/MontFerret/ferret/pkg/runtime/core"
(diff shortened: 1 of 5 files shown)


pkg/stdlib/io/fs/write.go
```diff
@@ -0,0 +1,154 @@
...
+		params, err = parseParams(args[2])
+		if err != nil {
+			return values.None, core.Error(
+				err,
+				"parse `params` argument",
+			)
+		}
+	}
+
+	flag, err := parseWriteMode(params.Mode)
```
ziflex: Why don't you parse it in ``parseParams`` function?
I think ``params`` struct should contain already parsed value.
What do you think about it?

3timeslazy: @ziflex good idea!

ziflex:
Awesome! Thanks for the PR!

Issue #397, opened 2019-10-13
error while trying to close a browser page.


gyy52380:
**Describe the bug**
When using the cdp driver, during closing of a browser page, this error sometimes appears.

```
{"level":"warn","time":"x","url":"x","error":"rpcc: the connection is closing","time":"x","message":"failed to close browser page"}
{"level":"error","time":"x","error":": rpcc: the connection is closing: session: detach timed out for session 5C391DF4E758E985AE3CBAA03774E562","time":"x","message":"closing root scope"}
```

**To Reproduce**
Run a single program 10-50 times, 2 - 4 instances in parallel. Observe the error (sometimes).

**Expected behavior**
No error should be printed.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome 77
 - Version: latest

**Additional context**
I only observed this issue when I had multiple programs running in parallel (2 to 4 instances at a time).

I'm not sure the library even needs to close the page manually. [Here](https://github.com/MontFerret/ferret/blob/master/pkg/drivers/cdp/page.go), on line 377 the page is closed manually. This looks weird because of the `context.Background()` passed in as the parameter. At the end of the function the page's connection is closed. I think that this should be enough to close the page.

laylwintthu396:
09782613844

solrac97gr:
I think this Issue is fixed, maybe someone need to close it?

PR #408, opened 2019-10-27
Bugfix for changes in PR #402

gyy52380:
This PR fixes a bug where runLoop would never end in some cases.

If a cdp connection is terminated due to network errors its event clients might not get closed properly (this might even be a bug in the cdp library, because this doesn't happen every time).

Adding a case which checks if the context has been canceled fixes this. I forgot to add this case in my initial PR. Sorry for that!

### DIFF
--- a/pkg/drivers/cdp/events/broker.go
+++ b/pkg/drivers/cdp/events/broker.go
@@ -191,6 +191,9 @@ func (broker *EventBroker) StopAndClose() error {
 func (broker *EventBroker) runLoop(ctx context.Context) {
 	for {
 		select {
+		case <-ctx.Done():
+			return
+
 		case <-broker.onLoad.Ready():
 			if ctxDone(ctx) {
 				return


ziflex:
Hey, thanks for your help!
But could you write some unit tests to avoid / detect problems in the future?

gyy52380:
How would you want me to unit test this? 

This bug is related to #397 and we don't have a solution for that yet.

This bug was previously handled by the `WithTimeout` context I removed in #407 , but it didn't handle it "correctly" -- you would just get mysterious "event context timed out" messages.

gyy52380:
I figured out what caused this bug.

When the chrome websocket connection was forcibly closed, the `cdp.Page` would die and [`broker.Close()`]{https://github.com/MontFerret/ferret/blob/master/pkg/drivers/cdp/events/broker.go#L161) would get called. Broker's clients wouldn't get closed properly because the connection was already closed. (here `*.Close()` would return an error, but we ignore it -- there is nothing we can really do with this information). That would cause their respective `Ready()` channels to not close which would block `runLoop()` indefinitely.

ziflex:
> How would you want me to unit test this?
> 
> This bug is related to #397 and we don't have a solution for that yet.
> 
> This bug was previously handled by the `WithTimeout` context I removed in #407 , but it didn't handle it "correctly" -- you would just get mysterious "event context timed out" messages.

There are already some unit tests for the broker in the repo. You can find them [here](https://github.com/MontFerret/ferret/blob/master/pkg/drivers/cdp/events/broker_test.go).
I was thinking of imitating these edge cases that you have covered in this and previous PRs.

ziflex:
> broker.Close()

What if we call ``.Close`` of an underlying document  before calling the broker's method? [Here](https://github.com/MontFerret/ferret/blob/master/pkg/drivers/cdp/page.go#L368) for example.

ziflex:
Ok, I will merge it for now, because I need the fix. 
I will add some unit tests.

gyy52380:
Sorry for not replying sooner. I'm currently on vacation and won't be contributing for a while. Thanks for merging the PR!

ziflex:
That's ok, have fun! :)

PR #453, opened 2020-02-18
USE keyword

3timeslazy:


pkg/parser/antlr/FqlParser.g4
```diff
@@ -6,8 +6,16 @@ program
...
+    : useExpressionAs
+    ;  
+
+useExpressionAs
+    : Use As
+    ;
+
 body
-    : (bodyStatement)* bodyExpression
+    : useExpression
```
ziflex: You need create ``head`` section. And add it to ``program``.

pkg/parser/antlr/FqlParser.g4
```diff
@@ -6,8 +6,16 @@ program
     : body
     ;
 
+useExpression
```
ziflex: Where is just ``USE``?

pkg/parser/antlr/FqlParser.g4
3timeslazy: It's works local now! Thanks!

PR #470, opened 2020-04-19
MVP USE statement

3timeslazy:


pkg/compiler/compiler_use_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/MontFerret/ferret/pkg/stdlib/strings"
+	"github.com/MontFerret/ferret/pkg/stdlib/types"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestUseExpression(t *testing.T) {
+	c := compiler.New()
+
+	c.Namespace("X").
+		RegisterFunction("INCLUDES", strings.Contains)
```
ziflex, line 17: Could you use a more unique name just to make sure that there is no a global function with such name?

pkg/compiler/compiler_use_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+				So(err, ShouldBeNil)
+				out := p.MustRun(context.Background())
+
+				So(string(out), ShouldEqual, `"true"`)
+			})
+
+			Convey("USE namespace twice", func() {
+				p, err := c.Compile(`
+				USE X
```
ziflex, line 56: Do you think we should allow this to compile?

pkg/compiler/visitor.go
```diff
@@ -50,6 +55,69 @@ func (v *visitor) VisitProgram(ctx *fql.ProgramContext) interface{} {
...
+	useexpr := head.UseExpression().(*fql.UseExpressionContext)
+
+	// TODO: Think about improving collision analysis to display more detailed errors.
+	// For example, "mamespaces X and Y both contain function F"
+	if iuse := useexpr.Use(); iuse != nil {
+		ns := iuse.(*fql.UseContext).
+			NamespaceIdentifier().
+			GetText()
+
+		err := deleteNamespace(v.funcs, ns)
```
ziflex, line 79: why are you removing it?
Can't I use something like this:
```
USE IO::FS

LET content = READ(@file)

IO::FS::WRITE(@anotherFile, content)

RETURN NONE
```

pkg/compiler/compiler_use_test.go
3timeslazy: Smth like `XXX_CONTAINS`?

pkg/compiler/compiler_use_test.go
3timeslazy: Think not. Will fix it

pkg/compiler/visitor.go
3timeslazy: I don’t see cases where this could be useful.  Maybe you have some examples?


I think this is a more clean approach. Always use with the prefix, or without.

pkg/compiler/compiler_use_test.go
ziflex: Yes

pkg/compiler/visitor.go
ziflex: It might be not very intuitive for people coming from another languages where both cases would work (Java, C#, Rust)

ziflex: Here is a use case. I'm working on an assertion library for Ferret and there are 2 functions that may cause some problems: ``T::TRUE`` and ``T::FALSE``.
If you don't allow to use full path, there will be a collision with global boolean values ``TRUE`` and ``FALSE``.

PR #471, opened 2020-04-19
Added scroll options

ziflex:


pkg/stdlib/html/lib.go
```diff
@@ -105,3 +105,43 @@ func waitTimeout(ctx context.Context, value values.Int) (context.Context, contex
 		time.Duration(value)*time.Millisecond,
 	)
 }
+
+func defaultScrollOptions() drivers.ScrollOptions {
+	return drivers.ScrollOptions{
+		Behavior: drivers.ScrollBehaviorInstant,
```
3timeslazy, line 111: We can set int value of `ScrollBehaviorInstant` to 0 instead 2. And do the same with `ScrollHorizontalAlignmentCenter` and `ScrollVerticalAlignmentCenter`.

As a result, method `defaultScrollOptions` will not be needed, an empty structure will already be the default value.

pkg/stdlib/html/scroll_element.go
```diff
@@ -11,44 +11,85 @@ import (
...
-	if len(args) == 2 {
-		doc, err := drivers.ToDocument(args[0])
+	var doc drivers.HTMLDocument
+	var el drivers.HTMLElement
+	var selector values.String
+	var opts drivers.ScrollOptions
+
+	if len(args) == 3 {
+		if err = core.ValidateType(args[1], types.String); err != nil {
+			return values.None, err
```
3timeslazy, line 30: Maybe would be better to wrap errors smth like `errors.Wrap(err, `validate "selector"`)`

pkg/stdlib/html/scroll_element.go
```diff
@@ -11,44 +11,85 @@ import (
...
-		return values.None, err
+		return values.None, doc.GetElement().ScrollIntoView(ctx, opts)
+	}
+
+	if el != nil {
+		return values.None, el.ScrollIntoView(ctx, opts)
 	}
 
-	return values.None, el.ScrollIntoView(ctx)
+	return values.None, core.TypeError(args[0].Type(), drivers.HTMLPageType, drivers.HTMLDocumentType, drivers.HTMLElementType)
```
3timeslazy, line 94: Too long line

pkg/drivers/scroll.go
```diff
@@ -0,0 +1,121 @@
+package drivers
+
+type ScrollBehavior int
+
+const (
+	ScrollBehaviorAuto ScrollBehavior = 0
```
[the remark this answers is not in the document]
ziflex: https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior

pkg/drivers/scroll.go
```diff
@@ -0,0 +1,121 @@
...
+
+	ScrollBehaviorSmooth ScrollBehavior = 1
+
+	ScrollBehaviorInstant ScrollBehavior = 2
+)
+
+func NewScrollBehavior(value string) ScrollBehavior {
+	switch value {
+	case "auto":
```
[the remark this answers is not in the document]
ziflex: which means it would have to jump through all other cases.

pkg/drivers/scroll.go
```diff
@@ -0,0 +1,121 @@
...
+const (
+	ScrollBehaviorAuto ScrollBehavior = 0
+
+	ScrollBehaviorSmooth ScrollBehavior = 1
+
+	ScrollBehaviorInstant ScrollBehavior = 2
+)
+
+func NewScrollBehavior(value string) ScrollBehavior {
+	switch value {
```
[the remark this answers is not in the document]
ziflex: Good point

pkg/stdlib/html/lib.go
ziflex: I set to default values as they are in DOM spec, but I guess you are right.

pkg/drivers/scroll.go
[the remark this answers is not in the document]
3timeslazy: I think write comment above `ScrollBehavior` containing this link is a good idea.

pkg/stdlib/html/scroll_bottom.go
```diff
@@ -23,5 +24,17 @@ func ScrollBottom(ctx context.Context, args ...core.Value) (core.Value, error) {
...
-	return values.None, doc.ScrollBottom(ctx)
+	var opts drivers.ScrollOptions
+
+	if len(args) > 1 {
+		opts, err = toScrollOptions(args[1])
+
+		if err != nil {
+			return values.None, err
+		}
+	} else {
```
3timeslazy, line 35: `opts` already is `drivers.ScrollOptions{}` before `if len(args) > 1 {`. Therefore, this branch can be removed.

pkg/stdlib/html/scroll_top.go
```diff
@@ -23,5 +24,17 @@ func ScrollTop(ctx context.Context, args ...core.Value) (core.Value, error) {
...
-	return values.None, doc.ScrollTop(ctx)
+	var opts drivers.ScrollOptions
+
+	if len(args) > 1 {
+		opts, err = toScrollOptions(args[1])
+
+		if err != nil {
+			return values.None, err
+		}
+	} else {
```
3timeslazy, line 35: `opts` already is `drivers.ScrollOptions{}` before `if len(args) > 1`. Therefore, this branch can be removed.

pkg/stdlib/html/scroll_xy.go
```diff
@@ -41,5 +42,17 @@ func ScrollXY(ctx context.Context, args ...core.Value) (core.Value, error) {
...
-	return values.None, doc.ScrollByXY(ctx, x, y)
+	var opts drivers.ScrollOptions
+
+	if len(args) > 3 {
+		opts, err = toScrollOptions(args[3])
+
+		if err != nil {
+			return values.None, err
+		}
+	} else {
```
3timeslazy, line 53: `opts` already is `drivers.ScrollOptions{}` before `if len(args) > 3`. Therefore, this branch can be removed.

pkg/drivers/scroll.go
```diff
@@ -0,0 +1,119 @@
+package drivers
+
+import "strings"
+
+type ScrollBehavior int
+
+const (
+	ScrollBehaviorInstant ScrollBehavior = 0
```
3timeslazy, line 8: Formatting different from 
```
const (
	ScrollVerticalAlignmentCenter  ScrollVerticalAlignment = 0
	ScrollVerticalAlignmentStart   ScrollVerticalAlignment = 1
	ScrollVerticalAlignmentEnd     ScrollVerticalAlignment = 2
	ScrollVerticalAlignmentNearest ScrollVerticalAlignment = 3
)

const (
	ScrollHorizontalAlignmentCenter  ScrollHorizontalAlignment = 0
	ScrollHorizontalAlignmentStart   ScrollHorizontalAlignment = 1
	ScrollHorizontalAlignmentEnd     ScrollHorizontalAlignment = 2
	ScrollHorizontalAlignmentNearest ScrollHorizontalAlignment = 3
)
```

Remove empty lines :)

PR #481, opened 2020-04-25
Added support of using keywords in namespaces

### LINKED ISSUE #475: Unable to use keywords as functions using namespaces
**Describe the bug**
There is no way to use reserved keywords along side with namespaces.

**To Reproduce**
Register a function ``TRUE`` within any namespace, and try to use it:
```
T::TRUE(false)
T::LIMIT(1)
```

**Expected behavior**
It must be possible to use keywords in such scenarios.

ziflex:
Closes #475


pkg/compiler/compiler_func_ns_test.go
```diff
@@ -78,4 +80,72 @@ func TestFunctionNSCall(t *testing.T) {
...
+			"Any",
+			"Aggregate",
+			"Like",
+			"Not",
+			"In",
+		}
+
+		for _, kw := range keywords {
+			segment := strings.ToUpper(kw)
+			err := c.Namespace("T").Namespace(segment).RegisterFunction("TEST", func(ctx context.Context, args ...core.Value) (core.Value, error) {
```
3timeslazy, line 117: maybe we should add package `internal/testutil` that contains stub like 
```go
func True(ctx context.Context, args ...core.Value) (core.Value, error) {
    return values.True, nil
}
```
to not rewriting it every time?

also line is too long :)

pkg/compiler/compiler_func_ns_test.go
```diff
@@ -78,4 +80,72 @@ func TestFunctionNSCall(t *testing.T) {
...
+				return values.True, nil
+			})
+
+			So(err, ShouldBeNil)
+
+			err = c.Namespace("T").Namespace(segment).RegisterFunction(segment, func(ctx context.Context, args ...core.Value) (core.Value, error) {
```
3timeslazy, line 123: see above

pkg/compiler/compiler_func_ns_test.go
```diff
@@ -78,4 +80,72 @@ func TestFunctionNSCall(t *testing.T) {
...
+			"All",
+			"Any",
+			"Aggregate",
+			"Like",
+			"Not",
+			"In",
+		}
+
+		for _, kw := range keywords {
+			segment := strings.ToUpper(kw)
```
3timeslazy, line 116: there is no need to do `strings.ToUpper` because `.RegisterFunction` doing it.

pkg/parser/antlr/FqlParser.g4
```diff
@@ -223,8 +223,37 @@ namespace
     : (NamespaceSegment)*
     ;
 
+functionIdentifier
```
3timeslazy, line 226: `Use`, `Or`, `And` is not a function identifiers. Maybe there is better name for it? `Keyword`?

pkg/compiler/compiler_func_ns_test.go
```diff
@@ -78,4 +80,72 @@ func TestFunctionNSCall(t *testing.T) {
 
 		So(err, ShouldNotBeNil)
 	})
+
+	Convey("Should use keywords", t, func() {
+		c := compiler.New()
+
+		keywords := []string{
```
3timeslazy, line 87: Is it working with `TRUE` and `FALSE`?

pkg/compiler/compiler_func_ns_test.go
ziflex: The function itself is not important here. It does not matter what it returns.

pkg/compiler/compiler_func_ns_test.go
ziflex: Same.

pkg/parser/antlr/FqlParser.g4
ziflex: ``TEST::USE()`` is a function :)

pkg/compiler/compiler_func_ns_test.go
ziflex: Yes

3timeslazy: In this case, I suggest adding cases with `TRUE` and `FALSE`

pkg/parser/antlr/FqlParser.g4
3timeslazy: I understand :) I missed `: Identifier` at the beginning and thought there was only a list of keywords.  Anyway, why `TRUE` and `FALSE` not on this list?)

ziflex: They are. ``BooleanLiteral``.

pkg/compiler/compiler_func_ns_test.go
ziflex: Ah, good catch. I need to "unwrap" ``BooleanLiteral`` and few other literals.

PR #501, opened 2020-05-09
Added fuzzer

AdamKorcz:
This PR adds a fuzzer that targets `Compile()`

I managed to get the fuzzer running on oss-fuzz's platform as well, and I propose setting ferret up on in oss-fuzz. This will allow oss-fuzz to run this fuzzer as well as future fuzzers continuously. If a bug is discovered, oss-fuzz sends a bug report via email to all maintainers on their contact list.
It doesn't cost anything to be a part of oss-fuzz, but the bugs do need to get fixed.

If you would like to setup ferret on oss-fuzz, I need a primary contact email address and the email addresses of all maintainers to add to the email list. This can be updated at anytime.

Google might ask about the userbase of ferret in order to accept the project, so if you have any information about any companies or other packages using ferret, it would help with getting ferret accepted.

### DIFF
--- a/pkg/compiler/fuzz/fuzz.go
+++ b/pkg/compiler/fuzz/fuzz.go
@@ -4,11 +4,14 @@ import (
 	"github.com/MontFerret/ferret/pkg/compiler"
 )
 
-// This function is our fuzzer
+// Below function is our fuzzer.
+// The fuzzer is run on oss-fuzz.
+// Link to the project on oss-fuzz
+// will be added once the project is up.
 func Fuzz(data []byte) int {
 	c := compiler.New()
 	p, err := c.Compile(string(data))
-	if err != nil || p == nil{
+	if err != nil || p == nil {
 		return 0
 	}
 	return 1
--- a/pkg/compiler/fuzz/fuzz.go
+++ b/pkg/compiler/fuzz/fuzz.go
@@ -4,7 +4,7 @@ import (
 	"github.com/MontFerret/ferret/pkg/compiler"
 )
 
-// Below function is our fuzzer.
+// Fuzz is our fuzzer.
 // The fuzzer is run on oss-fuzz.
 // Link to the project on oss-fuzz
 // will be added once the project is up.
--- a/pkg/compiler/fuzz/fuzz.go
+++ b/pkg/compiler/fuzz/fuzz.go
@@ -1,0 +1,14 @@
+package compiler_test
+
+import (
+	"github.com/MontFerret/ferret/pkg/compiler"
+)
+
+func Fuzz(data []byte) int {
+	c := compiler.New()
+	p, err := c.Compile(string(data))
+	if err!=nil || p==nil{
+		return 0
+	}
+	return 1
+}
--- a/pkg/compiler/fuzz/fuzz.go
+++ b/pkg/compiler/fuzz/fuzz.go
@@ -4,10 +4,11 @@ import (
 	"github.com/MontFerret/ferret/pkg/compiler"
 )
 
+// This function is our fuzzer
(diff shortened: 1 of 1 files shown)


ziflex:
Hey, Adam, that's a great idea, thanks for offering!
The Ferret team would like to participate in this. Where should I leave the information?

AdamKorcz:
@ziflex Awesome! Feel free to leave the email address here, or if you prefer, please send it to me at adam /at/ adalogics /dot/ com.

pkg/compiler/fuzz/fuzz.go
```diff
@@ -0,0 +1,14 @@
+package compiler_test
+
+import (
+	"github.com/MontFerret/ferret/pkg/compiler"
+)
+
+func Fuzz(data []byte) int {
+	c := compiler.New()
+	p, err := c.Compile(string(data))
+	if err!=nil || p==nil{
```
3timeslazy, line 10: variables and operators should be separated by space. So, line should be `if err != nil || p == nil`

pkg/compiler/fuzz/fuzz.go
```diff
@@ -0,0 +1,14 @@
+package compiler_test
+
+import (
+	"github.com/MontFerret/ferret/pkg/compiler"
+)
+
+func Fuzz(data []byte) int {
```
3timeslazy, line 7: I think there should be a comment explaining why this function is needed.

AdamKorcz:
@3timeslazy This is not needed on this side. These commands are needed to setup the fuzzers on their systems. The fuzzer uploaded is what is needed on upstream repository to run the fuzzer.

pkg/compiler/fuzz/fuzz.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/MontFerret/ferret/pkg/compiler"
+)
+
+// This function is our fuzzer
+func Fuzz(data []byte) int {
+	c := compiler.New()
+	p, err := c.Compile(string(data))
+	if err != nil || p == nil{
```
3timeslazy, line 11: Linter falls here https://travis-ci.com/github/MontFerret/ferret/jobs/331199355

pkg/compiler/fuzz/fuzz.go
```diff
@@ -0,0 +1,15 @@
+package compiler_test
+
+import (
+	"github.com/MontFerret/ferret/pkg/compiler"
+)
+
+// This function is our fuzzer
```
3timeslazy, line 7: This comment is not descriptive enough and does not satisfy Go's code style. Write something like
```go
// Fuzz is a fuzzer for oss-fuzz platform.
//
// Read more about it here: https://github.com/google/oss-fuzz
func Fuzz(data []byte) int {}
```

3timeslazy:
@AdamKorcz thanks for explanation! 

P.S.
Added a few comments :)

AdamKorcz:
@3timeslazy Thanks for the comments. Regarding the in-code comment: The fuzzer is not specific to oss-fuzz and it can be run outside of oss-fuzz as well. Hence I suggest to not make the description of it too specific to oss-fuzz.

3timeslazy:
@AdamKorcz you're right about a too specific comment. 

And the last comment :) Replace `Below function` with `Fuzz` pls. This is a code style requirement. Comment should begin with the function name.

Source: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

ziflex:
Thank you all for the PR!
Keep fingers crossed that the project gets approval.

AdamKorcz:
Thank you from here as well for all the good comments on a sunday. Crossing my fingers!

PR #513, opened 2020-05-20
Generate doc YAML representation for Ferret stdlib

ksdme:
- [x] Use tokens which compiler exposes instead of the implementation name.


ksdme:
Generated representation for current stdlib: https://gist.github.com/ksdme/b8e77c52bfac4633b7a8996c2e2f1710

The plan I have is to first generate an internal rep like this and hook up Hugo to take this YAML and build out the pages.

I am not sure if the scripts belong to this repo or if they should be moved elsewhere, I think it depends on a few things,

1. Would it bloat the repo?
2. Would docs be auto-generated? If so, how would the CI work?

ksdme:
CC @ziflex @3timeslazy

pkg/stdlib/arrays/first.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // First returns a first element from a given array.
 // @param arr (Array) - Target array.
-// @returns element (Read) - First element in a given array.
+// @returns (Read) - First element in a given array.
```
ziflex, line 13: ``(Value))`` I believe.

pkg/stdlib/strings/json.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // JSONParse returns a FQL value described by the JSON-encoded input string.
 // @params text (String) - The string to parse as JSON.
-// @returns FQL value (Read)
+// @returns (Read) - Returns FQL Value
```
ziflex, line 13: ``(Value)``.
I think ``GoLand`` renamed it to ``Read`` during one of refactorings :)

ziflex:
First of all, it looks really great!

> Generated representation for current stdlib: https://gist.github.com/ksdme/b8e77c52bfac4633b7a8996c2e2f1710
> 
> The plan I have is to first generate an internal rep like this and hook up Hugo to take this YAML and build out the pages.
> 
> I am not sure if the scripts belong to this repo or if they should be moved elsewhere, I think it depends on a few things,
> 
> 1. Would it bloat the repo?
> 2. Would docs be auto-generated? If so, how would the CI work?

Second, this is what I was thinking about. Because, eventually, we will use this tool in [this repo](https://github.com/MontFerret/montferret.github.io).

ksdme:
Updated representation https://gist.github.com/ksdme/b8e77c52bfac4633b7a8996c2e2f1710

ksdme:
Function signature pattern needs to be independent of argument names. https://github.com/MontFerret/ferret/blob/master/pkg/stdlib/collections/length.go#L12 is failing as of now.

ksdme:
https://github.com/MontFerret/montferret.github.io/pull/12

ziflex:
Fantastic job!
But let's move the tool [Here](https://github.com/MontFerret/docgen) in order to keep the repo slim.

ksdme:
https://github.com/MontFerret/docgen/pull/1 moves the actual rep generation scripts to docgen.

ksdme:
@ziflex I should've mentioned in the other comment that the changes in this PR still need to be merged, they fix and standardize comment structure along with some related refactoring.

PR #523, opened 2020-05-29
Fix an emoji error in the string.

leolovenet:
strconv.Unquote will cause garbled code when the string contains emoticons and should be left to the default json.Unmarshal to handle

### DIFF
--- a/pkg/drivers/cdp/eval/helpers.go
+++ b/pkg/drivers/cdp/eval/helpers.go
@@ -2,7 +2,6 @@ package eval
 
 import (
 	"fmt"
-	"strconv"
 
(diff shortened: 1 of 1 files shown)


ziflex:
Hey thanks for the PR.
Could you provide some tests to demonstrate what issue gets solved?

leolovenet:
Yea, sorry for the missing, Here's a test script, I found the problem, impact only under cdp driver
```
LET page = DOCUMENT("https://getemoji.com/", {
    driver: "cdp"
})

RETURN ELEMENT(page, "body > div.container > div > p:nth-child(9)").innerText
```
Xnip2020-06-02_00-58-44

Issue #518, opened 2020-05-24
CDP driver does not close tabs


ziflex:
Whenever CLI process gets interrupted or killed, CDP driver must (and used to) close all open tabs.
It stopped doing this.

Pangoraw:
I am interested in solving this one. It should close tabs when handling signals like SIGTERM sent with `kill -TERM <PID>` ? 
For me when running CTRL+C the tabs close.

ziflex:
> I am interested in solving this one. It should close tabs when handling signals like SIGTERM sent with `kill -TERM <PID>` ?
> For me when running CTRL+C the tabs close.

Hmm... I might spend more time on this but when I was working on [lab](https://github.com/MontFerret/lab) I have noticed when I interrupted execution, tabs remained open. Then I tested it via REPL and got same effect.

PR #524, opened 2020-05-29
allow context cancellation in WAIT

Pangoraw:
While investigating issue https://github.com/MontFerret/ferret/issues/518, I noticed that when pressing CTRL+C during `WAIT()` would not cause the program to stop instantly, but only at the end of the duration. You can now press CTRL+C when waiting and the stopping will happen instantly.

NOTE: I am not sure about the error message and whether or not I should add a test.

### DIFF
--- a/pkg/stdlib/utils/wait.go
+++ b/pkg/stdlib/utils/wait.go
@@ -10,7 +10,7 @@ import (
 
 // Wait pauses the execution for a given period.
 // @param timeout (Float|Int) - Number value which indicates for how long to stop an execution.
-func Wait(_ context.Context, args ...core.Value) (core.Value, error) {
+func Wait(ctx context.Context, args ...core.Value) (core.Value, error) {
 	err := core.ValidateArgs(args, 1, 1)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/stdlib/utils/wait.go
```diff
@@ -19,7 +20,13 @@ func Wait(_ context.Context, args ...core.Value) (core.Value, error) {
 
 	arg := values.ToInt(args[0])
 
-	time.Sleep(time.Millisecond * time.Duration(arg))
+	timer := time.NewTimer(time.Millisecond * time.Duration(arg))
+	select {
+	case <-ctx.Done():
+		timer.Stop()
+		return values.None, errors.New("context cancelled")
```
ziflex, line 27: you can return ``ctx.Err()``

ziflex:
Good catch!

Pangoraw:
[This test](https://github.com/MontFerret/ferret/blob/master/pkg/runtime/program_test.go#L31) started to fail randomly because the error message was changing from either `core.ErrTerminated` to `context canceled: WAIT(1000)`. So in order to make sure the context was already canceled, I cancel it before running the program (see https://github.com/MontFerret/ferret/pull/524/commits/70650cde2b007a45e1638202fc16bc8f784d3411).

PR #531, opened 2020-06-18
Removed caching and property tracking

ziflex:
- Removed attrs change handlers
- Removed caching

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,7 @@ os:
 - linux
 
 go:
-- "1.11.x"
-- "1.12.x"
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -6,8 +6,7 @@ os:
 - linux
 
 go:
-- "1.11.x"
-- "1.12.x"
+- "1.13.x"
```
3timeslazy, line 9: go1.14 already released!

ziflex: It's "stable", isn't it?

Issue #488, opened 2020-04-29
How to get rid of converted characters in URLs 


asitemade4u:
I am developing a crawler and so far, so very good: thank you for this outstanding crawler.

The only issue is that, in the returned URLs, there is a `&` character which gets converted into `\u0026`, thus: `"https://thedomain/alphabet=M\u0026borough=Bronx"`

So I tried to replace it, either by using `SUBSTITUTE`:
```RETURN SUBSTITUTE(prfx + letter.attributes.href, "\u0026", "&")```

or `REGEX_REPLACE`.

In both cases, the `\u0026` string is NOT replaced and remains embedded into the resulting URLs.
However, when I try `SUBSTITUTE`say on `a` -> `z` it works fine.

Is it a limitation of JSON, which I use as an output format?
How can I get rid of the converted string as it prevents me from crawling at the lower levels of the website.

**Comment by ziflex** [comment]
Thank you and I'm glad that you like it!

I assume you are using ``cdp`` driver, right? If so, I think the problem is that CDP is encodes urls before sending it to Ferret since it's a JSON-based communication.
``DECODE_URI_COMPONENT`` is supposed to solve the problem but it seems not working:
```
Welcome to Ferret REPL
Please use `exit` or `Ctrl-D` to exit this program.
> RETURN DECODE_URI_COMPONENT("https://thedomain/alphabet=M\u0026borough=Bronx")
"https://thedomain/alphabet=M\\u0026borough=Bronx"
```

Therefore I mark it as a bug.

asitemade4u:
Thank you for the response!

I have a related 2question as you just mentioned cdp.
I would like not to use it but I have not found a way to access the 
server which I do NOT host locally, but on a "real, physical" local 
server, under docker.

So, is there a way to connect to a remote server without resorting to CDP?

Currently, when executing a script, I do: ferret --cdp http://<the 
server ip>:9222 <thefile>.aql

Please also note I am using the AQL extension because the only VS Code 
plugin that formats and lints the language does not work with FQL, only AQL.

ziflex:
Not sure I fully understand you.
CDP driver uses Chrome to communicate with web pages and you need it only if your target page is dynamically rendered and/or requires some user interaction to retrieve data you need.

asitemade4u:
Sorry for being so obscure.

The question was: how do I connect to a remote Ferret server, hosted on 
a different IP address than localhost, from the cli?

Something like:

[...]

PR #559, opened 2020-10-03
JSON encoding with no HTML escaping

### LINKED ISSUE #550: DECODE_URI_COMPONENT seems broken
like discussed in this issue : https://github.com/MontFerret/ferret/issues/488

```
Welcome to Ferret REPL
Please use `exit` or `Ctrl-D` to exit this program.
> RETURN DECODE_URI_COMPONENT("https://thedomain/alphabet=M\u0026borough=Bronx")
"https://thedomain/alphabet=M\\u0026borough=Bronx"
```

d2bit:
Hi,
As far as I know the problem is that we are encoding as JSON the final result 
https://github.com/MontFerret/ferret/blob/master/pkg/runtime/program.go#L103
So that, being the HTML special chars escaping enabled by default in the `encoding/json` package it encoded HTML escaped characters as `"&"` to `"\\u0026"` (notice the double `\`, as they are 6 runes, instead of 1)

So that, my suggestion is not to escape HTML characters.

For that, I've created an encoder package and updated all the calls to `json.Marshall` in `runtime/values` to use the new encoder package.

I have not much experience in Go, I've tried to mimic the project structure for adding the new package but maybe you could suggest a better place/names for that. 

Fixes #550 and #488

### DIFF
--- a/pkg/runtime/values/array.go
+++ b/pkg/runtime/values/array.go
@@ -3,11 +3,11 @@ package values
 import (
 	"context"
 	"encoding/binary"
-	"encoding/json"
 	"hash/fnv"
 	"sort"
 
 	"github.com/MontFerret/ferret/pkg/runtime/core"
+	"github.com/MontFerret/ferret/pkg/runtime/values/encoder"
 	"github.com/MontFerret/ferret/pkg/runtime/values/types"
 )
 
@@ -30,7 +30,7 @@ func NewArrayWith(values ...core.Value) *Array {
 }
 
 func (t *Array) MarshalJSON() ([]byte, error) {
-	return json.Marshal(t.items)
+	return encoder.EncodeJSON(t.items)
 }
 
 func (t *Array) Type() core.Type {
--- a/pkg/runtime/values/boolean.go
+++ b/pkg/runtime/values/boolean.go
@@ -1,11 +1,11 @@
 package values
 
 import (
-	"encoding/json"
 	"hash/fnv"
 	"strings"
 
 	"github.com/MontFerret/ferret/pkg/runtime/core"
+	"github.com/MontFerret/ferret/pkg/runtime/values/encoder"
 	"github.com/MontFerret/ferret/pkg/runtime/values/types"
 )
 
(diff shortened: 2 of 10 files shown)


ziflex:
Hey, sorry for the late review. 
But won't it produce an invalid JSON string?

d2bit:
> But won't it produce an invalid JSON string?

When I've tested it locally it worked just fine.

ziflex:
Hey, sorry again for so long delay.
I'm afraid that a shared encoder may cause same issues when we have multiple programs running in parallel since they are not thread safe. 
More on this [here](https://github.com/golang/go/issues/32957)

ziflex:
Alternative way could be this: since all values are internal and they do not leak outside from the runtime, we could deviate from canonical ``Marshaller`` interface. The solution might look something like this:

```go
type Encoder interface {
    Encode(value interface{}) ([]byte, error)
}

type Marshaler interface {
    Marshal(encoder Encoder) ([]byte, error)
}

type Value interface {
    Marshaller
    Type() Type
    String() string
    Compare(other Value) int64
    Unwrap() interface{}
    Hash() uint64
    Copy() Value
}
```

So, that would allow us to create a separate encoder that would be passed in during results serialization. (We could even pass a custom one, to encode as a binary data, for example)

ziflex:
> Alternative way could be this: since all values are internal and they do not leak outside from the runtime, we could deviate from canonical `Marshaller` interface. The solution might look something like this:
> 
> ```go
> type Encoder interface {
>     Encode(value interface{}) ([]byte, error)
> }
> 
> type Marshaler interface {
>     Marshal(encoder Encoder) ([]byte, error)
> }
> 
> type Value interface {
>     Marshaller
>     Type() Type
>     String() string
>     Compare(other Value) int64
>     Unwrap() interface{}
>     Hash() uint64
>     Copy() Value
> }
> ```
> 
> So, that would allow us to create a separate encoder that would be passed in during results serialization. (We could even pass a custom one, to encode as a binary data, for example)

Ok, I think it's not a good idea to deviate from the Go's Marshaler interface because it will lead us to the need of creating our own JSON encoder, the task I do not want to do.

I think we might do it easier by replacing use of ``encoding/json`` package by ``github.com/wI2L/jettison``:

```go
jettison.MarshalOpt(val, jettison.NoHTMLEscaping())
```

ziflex:
Closing in favor of #573

Issue #483, opened 2020-04-27
RAND() always same result


Devilmaydante:
Hello,

I have a pb with the RAND function
I use this piece of code :

> LET randomValue = RAND()
> 
> RETURN {
> 	A:randomValue
> }

Each time I execute this code, I obtain the same value (0.544155573000885}

 - OS: Windows 10
 - Browser Chromium,
 - Version 84.0.4129.0 

Thank for your help

PR #579, opened 2020-12-14
Add rand seed in RAND(min, max) #578

### LINKED ISSUE #578: RAND(0,100) always same result
Hey @u5surf  @ziflex 

since https://github.com/MontFerret/ferret/issues/483  and https://github.com/MontFerret/ferret/commit/7d7253325e7e465f16c47469b777799931987a8a RAND() seems to be ok 
but RAND(0,100) always return the same value also

Thank for everything

u5surf:
Fixes #578 
It had returned the same value for same reason as #483.
Because Random in pkg/runtime/core/helpers.go didn't change rand.Seed every time.

### DIFF
--- a/pkg/runtime/core/helpers.go
+++ b/pkg/runtime/core/helpers.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"math/rand"
 	"reflect"
+	"time"
 )
 
 func IsNil(input interface{}) bool {
@@ -46,6 +47,7 @@ func NumberLowerBoundary(input float64) float64 {
 }
 
 func Random(max float64, min float64) float64 {
+	rand.Seed(time.Now().UnixNano())
 	r := rand.Float64()
 	i := r * (max - min + 1)
 	out := math.Floor(i) + min


pkg/runtime/core/helpers.go
```diff
@@ -46,6 +47,7 @@ func NumberLowerBoundary(input float64) float64 {
 }
 
 func Random(max float64, min float64) float64 {
+	rand.Seed(time.Now().UnixNano())
```
ziflex, line 50: Can we create a new source instead?

ziflex: Create a new soruce and use this function from stdlib.

u5surf: @ziflex 
Thanks reviewing. your advice makes me confused a little.
First, why `core.Random` cannot be modified directly?
Indeed, it is used for a public function, but it doesn’t call any rand.Seed, thus `rand` has returned same value. Generally, it is wrong way that is getting same values everytime without changing the seed of the pesudo random  except using for testcase which is not allowed different value returns.

ziflex: Sorry, for the confusion. My point was that these functions do pretty much the same, so we could just have 1 implementation, instead of 2.

Second thing that I would like to address is creating a new seed source on each operation to make these function thread safe and give different results every time they are called.

ziflex: After reviewing it one more time with fresh pair of eyes, my comments above can be ignored :)

PR #585, opened 2021-01-21
custom http transport

bundleman:
I added option WithCustomTransport for http driver and  did refactoring code which create pester.Client. 

Added WithCustomTransport was necessary for control flag InsecureSkipVerify in section TLSClientConfig inside http.Transport.

PR #586, opened 2021-01-22
custom http transport

bundleman:
I added option WithCustomTransport for http driver and did refactoring code which create pester.Client.

Added WithCustomTransport was necessary for control flag InsecureSkipVerify in section TLSClientConfig inside http.Transport.

p.s. it is PR after review https://github.com/MontFerret/ferret/pull/585#issuecomment-764862377


ziflex:
Thanks for your PR!

PR #587, opened 2021-01-22
Check Kind() by 0

bundleman:
In `pkg/runtime/values/helpers.go` added Kind check to 0.
Otherwise, passing parameters with a nested nil structure leads to panic.


ziflex:
Good catch! 
Thank you for your contribution!

PR #608, opened 2021-04-01
Removed old CLI

### LINKED ISSUE #266: Split CLI and runtime packages
As the project grows, it's getting more messy to keep everything in one place.

The proposal is to split the project into following separate repos:

- [ ] ``ferret`` will contain only Ferret itself - parser, compiler, runtime and stdlib
- [x] ``cli`` will contain CLI.

Any thoughts?

ziflex:
Closes #266


pkg/drivers/container.go
```diff
@@ -0,0 +1,58 @@
...
+	drivers map[string]Driver
+}
+
+func NewContainer() *Container {
+	return &Container{
+		drivers: map[string]Driver{},
+	}
+}
+
+func (c *Container) HasDriver(name string) bool {
```
3timeslazy, line 17: I suggest to remove `Driver` word from `Container's` methods. The `Container` is in `drivers` package. Thus when we write `Container.Has` we know it means `has driver`

That's how `Container` calls looks in `ferret.go`:
```go
// drivers 
for _, drv := range i.drivers.RegisteredDrivers() { ... }
```
How will it look like after remove:
```go
for _, drv := range i.drivers.Registered() { ... }
```

In my opinion, the second code looks better

e2e/cli.go
```diff
@@ -0,0 +1,146 @@
...
+	if (stat.Mode() & os.ModeCharDevice) == 0 {
+		// check whether the app is getting a query via standard input
+		std := bufio.NewReader(os.Stdin)
+
+		b, err := ioutil.ReadAll(std)
+
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+			return
```
3timeslazy, line 90: Can be removed because of `os.Exit(1)`

e2e/cli.go
```diff
@@ -0,0 +1,146 @@
...
+
+		query = string(b)
+	} else if flag.NArg() > 0 {
+		// backward compatibility
+		content, err := ioutil.ReadFile(flag.Arg(0))
+
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+			return
```
3timeslazy, line 101: Can be removed because of `os.Exit(1)`

3timeslazy (changes requested):

pkg/drivers/container.go
ziflex: I wanted to this way originally, but then I saw that ``Namespace`` interface has ``*Function``, so to be consistent I decided to follow the pattern.

3timeslazy: I think it will be consistent because `namespace` and `function` are different entities. It's ok to write `Namespace.RegisteredFunctions()` and `Drivers.Registered` together

PR #609, opened 2021-04-05
setting Input charset

bundleman:
I am doing scraping in Russian segment. And i ran into a problem with encoding some pages at some sites. The pages have not only < meta charset="windows-1251"/> but also content have encoding at winfows-1251.
But goquery decoding page to utf-8 without own content encoding and the output i'm getting unreadable content. 
So, I added param “charset” (may by this need to other name – to reviewer notice) to document params and function decoding to utf-8 with given the “charset”.


ziflex:
How does Chrome handle such pages ? Do we need to make any changes in CDP driver?

bundleman:
> How does Chrome handle such pages ? Do we need to make any changes in CDP driver?

No, chrome handles that pages good.

PR #635, opened 2021-07-16
Fix cpu leakage

bundleman:


pkg/drivers/cdp/page.go
```diff
@@ -107,13 +105,10 @@ func LoadHTMLPage(
 		mouse,
 		keyboard,
 	)
-
 	if err != nil {
 		return nil, err
 	}
 
-	closers = append(closers, domManager)
```
bundleman, line 115: Sorry, there is a past experiment left

ziflex (approved):

PR #673, opened 2021-09-20
Feature/variables optimization

ziflex:
- Loops containing ``RETURN NONE`` return empty arrays now (almost zero-allocation)
- Variables having ``_`` name are ignored and do not get allocated to scope/stack.


pkg/runtime/expressions/for_result.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (f *ForResult) Push(value core.Value) {
+	if f.passThrough {
+		return
+	}
+
+	if f.distinct {
+		// We need to check whether the value already exists in the result set
+		hash := value.Hash()
+		_, exists := f.hashTable[hash]
```
3timeslazy, line 55: `f.hashTable` has type `map[uint64]bool` and value always `true`, so this can be rewritten as 
```go
if f.hashTable[hash] {
    return
}
```

3timeslazy: If the key does not exist `f.hashTable[hash]` returns default value for bool which is `false`

PR #678, opened 2021-10-06
Support colors on Windows

mattn:
image


ziflex:
Hey, thanks for your PR, but this CLI is for e2e testing only.
The main CLI has moved [here](https://github.com/MontFerret/cli).

PR #719, opened 2021-12-28
Many changes

bundleman:
1. Added options for http driver: WithTimeout, WithBodyLimit.
2. Changed default level from Debug to Trace when setting http headers. 
3. Added ferret function to_binary for convert data to array bytes (it's necessary for requests with body).
4. Added mutex to map with [params](https://github.com/MontFerret/ferret/commit/a849a8100c2b0417f2cf16bfc1a2a890e086a165) because in some cases, emerging race condition.
5. Big changes - it's unification using function document without DOM tree prepare. Current realization (IO::NET::HTTP) not using proxy and not retuned headers and cookies and duplicates code in http.Driver. So I rewrote it with HTTPResponse as returned object from ferret - [realization](https://github.com/MontFerret/ferret/commit/621233c47a1c32ebbd7c61993cd748542c55dd90) and [tests](https://github.com/MontFerret/ferret/commit/59deed3de16b0b29f526b5c6b1ba2ee595a07377)


ziflex:
oh, wow! Could you split them into several PR's, please?

ziflex:
Regarding №5 - I didn't quite understand the purpose of this method. Can't you achieve the same with ``IO::NET::HTTP::GET`` and ``PARSE`` methods?

ziflex (changes requested):
Please, split into separate PRs, since I have doubts regarding some of the features provided by this PR.

bundleman:
> Please, split into separate PRs, since I have doubts regarding some of the features provided by this PR.

Ok

bundleman:
> Regarding №5 - I didn't quite understand the purpose of this method. Can't you achieve the same with `IO::NET::HTTP::GET` and `PARSE` methods?

Because between document and IO::NET::HTTP no differences with the exception of using goquery. But`IO::NET::HTTP` is wordy and hasn't the usual signature as like `document`. So seems if using `document` it will look more unified.

ziflex:
You can always use ``USE`` to  avoid typing long namespace names.

Sorry but I'm not convicted that it's a good reason for adding a new method to the driver interface since you can accomplish the same with other existing API.

bundleman:
> You can always use `USE` to avoid typing long namespace names.
> 
> Sorry but I'm not convicted that it's a good reason for adding a new method to the driver interface since you can accomplish the same with other existing API.

If will use `US` it gets even more wordy. If you have many fqls it's problem - each time write `US` near IO::NET::HTTP and also you must remember about other signature where `url` inside object and cookies must not be array with objects but something, etc. This is another place for mistakes if you use this seldom.

> you can accomplish the same with other existing API
Which for example ? Current realization doesn't support returning cookies and using proxy. Or maybe I don't understand something? Or you suggest rewrite `IO::NET::HTTP`(I will remind this code duplicates code in driver.Open()) as i did it using `document` with `simpleRequest` ?

ziflex:
> > You can always use `USE` to avoid typing long namespace names.
> > Sorry but I'm not convicted that it's a good reason for adding a new method to the driver interface since you can accomplish the same with other existing API.
> 
> If will use `US` it gets even more wordy. If you have many fqls it's problem - each time write `US` near IO::NET::HTTP and also you must remember about other signature where `url` inside object and cookies must not be array with objects but something else, etc. This is another place for mistakes.

I don't see extra verbosity as a big problem.  It's just a fast copy pase. Or you can write a simple transformer that adds it to the top of your queries.

> 
> > you can accomplish the same with other existing API
> 
> Which for example ? Current realization doesn't support returning cookies and using proxy. Or maybe I don't understand something? Or you suggest rewrite `IO::NET::HTTP`(I will remind this code duplicates code in driver.Open()) as i did it using `document` with `simpleRequest` ?

What I'm suggesting is to use an existing API and keep single responsibility in place. Your changes make API less clear and add extra functionality that is not needed there.
I'm more open to extending HTTP API methods if there is inconsistency or lack of functionality. But not HTML drivers API. 
What you are trying to do is basically 2 function calls of the existing API - ``IO::NET::HTTP::REQUEST`` and ``PARSE``. The former returns raw body and that latter invokes a driver's parse method and returns a DOM tree.

bundleman:
> > > You can always use `USE` to avoid typing long namespace names.
> > > Sorry but I'm not convicted that it's a good reason for adding a new method to the driver interface since you can accomplish the same with other existing API.
> > 
> > 
> > If will use `US` it gets even more wordy. If you have many fqls it's problem - each time write `US` near IO::NET::HTTP and also you must remember about other signature where `url` inside object and cookies must not be array with objects but something else, etc. This is another place for mistakes.
> 
> I don't see extra verbosity as a big problem. It's just a fast copy pase. Or you can write a simple transformer that adds it to the top of your queries.
> 
> > > you can accomplish the same with other existing API

For what? This is additional work. 
And overall, this design doesn't look organic compared to the rest of the keywords - isn't that obvious?

> > Which for example ? Current realization doesn't support returning cookies and using proxy. Or maybe I don't understand something? Or you suggest rewrite `IO::NET::HTTP`(I will remind this code duplicates code in driver.Open()) as i did it using `document` with `simpleRequest` ?
> 
> What I'm suggesting is to use an existing API and keep single responsibility in place. Your changes make API less clear and add extra functionality that is not needed there. I'm more open to extending HTTP API methods if there is inconsistency or lack of functionality. But not HTML drivers API. What you are trying to do is basically 2 function calls of the existing API - `IO::NET::HTTP::REQUEST` and `PARSE`. The former returns raw body and that latter invokes a driver's parse method and returns a DOM tree.

These changes are not in the `HTML` driver (there is no such thing), but in the `HTTP`. In the api part, nothing changes - for the user the function remains the same, but the result may be different (for example, `xpath` can return different types and this is convenient), because the library allows you to do this, and this is great.

> What you are trying to do is basically 2 function calls of the existing API - `IO::NET::HTTP::REQUEST` and `PARSE`. The former returns raw body and that latter invokes a driver's parse method and returns a DOM tree.

No, I'm not trying to do this - the HTTPResponse is returned and not the HTMLDocument as `Parse` does, the method does not use `Parse` because there is no need for it.

p.s. If now I make my changes to IO::NET::HTTP, then there will be no backward compatibility. Users will have to rewrite their fqls.

ziflex:
> > > > You can always use `USE` to avoid typing long namespace names.
> > > > Sorry but I'm not convicted that it's a good reason for adding a new method to the driver interface since you can accomplish the same with other existing API.
> > > 
> > > 
> > > If will use `US` it gets even more wordy. If you have many fqls it's problem - each time write `US` near IO::NET::HTTP and also you must remember about other signature where `url` inside object and cookies must not be array with objects but something else, etc. This is another place for mistakes.
> > 
> > 
> > I don't see extra verbosity as a big problem. It's just a fast copy pase. Or you can write a simple transformer that adds it to the top of your queries.
> > > > you can accomplish the same with other existing API
> 
> For what? This is additional work. And overall, this design doesn't look organic compared to the rest of the keywords - isn't that obvious? Also, function signatures(`document` and `IO::NET::HTTP`) are different for the same actions.

This is the root of the misunderstanding. The functions are _not_ for the same actions. And they do _not_ do the same actions.
``DOCUMENT`` relies on registered HTTP drivers (``CDP`` and ``HTTP``, yes, poor naming choice) and is designed to provide extensible API for dealing with HTML objects. ``DOCUMENT`` is _not_ intended to make low-level HTTP requests. Period.

> 
> > > Which for example ? Current realization doesn't support returning cookies and using proxy. Or maybe I don't understand something? Or you suggest rewrite `IO::NET::HTTP`(I will remind this code duplicates code in driver.Open()) as i did it using `document` with `simpleRequest` ?
> > 
> > 
> > What I'm suggesting is to use an existing API and keep single responsibility in place. Your changes make API less clear and add extra functionality that is not needed there. I'm more open to extending HTTP API methods if there is inconsistency or lack of functionality. But not HTML drivers API. What you are trying to do is basically 2 function calls of the existing API - `IO::NET::HTTP::REQUEST` and `PARSE`. The former returns raw body and that latter invokes a driver's parse method and returns a DOM tree.
> 
> These changes are not in the `HTML` driver (there is no such thing), but in the `HTTP`. In the API part, nothing changes - for the user the function remains the same, but the result may be different (for example, `xpath` can return different types and this is convenient), because the library allows you to do this, and this is great.

``HTML`` driver is a general term describing an interface in the ``drivers`` module.

> > What you are trying to do is basically 2 function calls of the existing API - `IO::NET::HTTP::REQUEST` and `PARSE`. The former returns raw body and that latter invokes a driver's parse method and returns a DOM tree.
> 
> No, I'm not trying to do this - the `HTTPResponse` is returned with raw body and not the `HTMLDocument` with DOM as `Parse` does, the method does not use `Parse` because there is no need for it.

Then I do not understand why you are trying to add low-level HTTP functionality to HTML-only components.

> p.s. If now I make my changes to IO::NET::HTTP, then there will be no backward compatibility. Users will have to rewrite their fqls.
You don't have to break backward compatibility.

Anyway, I'm not willing to accept the changes in the driver. You always can register your own driver and use it as a default one. It's one of the design goals since the library can not cover each user's edge case.

ziflex:
Let's bring in other changes first, and then we'll figure out what to do with the HTTP thing.

bundleman:
> > > > > You can always use `USE` to avoid typing long namespace names.
> > > > > Sorry but I'm not convicted that it's a good reason for adding a new method to the driver interface since you can accomplish the same with other existing API.
> > > > 
> > > > 
> > > > If will use `US` it gets even more wordy. If you have many fqls it's problem - each time write `US` near IO::NET::HTTP and also you must remember about other signature where `url` inside object and cookies must not be array with objects but something else, etc. This is another place for mistakes.
> > > 
> > > 
> > > I don't see extra verbosity as a big problem. It's just a fast copy pase. Or you can write a simple transformer that adds it to the top of your queries.
> > > > > you can accomplish the same with other existing API
> > 
> > 
> > For what? This is additional work. And overall, this design doesn't look organic compared to the rest of the keywords - isn't that obvious? Also, function signatures(`document` and `IO::NET::HTTP`) are different for the same actions.
> 
> This is the root of our misunderstanding. The functions are _not_ for the same actions. And they do _not_ make the same actions. `DOCUMENT` relies on registered HTTP drivers (`CDP` and `HTTP`, and yes, poor naming choice) and is designed to provide extensible API for dealing with HTML objects. `DOCUMENT` is _not_ intended to make low-level HTTP requests. Period.
> 
> > > > Which for example ? Current realization doesn't support returning cookies and using proxy. Or maybe I don't understand something? Or you suggest rewrite `IO::NET::HTTP`(I will remind this code duplicates code in driver.Open()) as i did it using `document` with `simpleRequest` ?
> > > 
> > > 
> > > What I'm suggesting is to use an existing API and keep single responsibility in place. Your changes make API less clear and add extra functionality that is not needed there. I'm more open to extending HTTP API methods if there is inconsistency or lack of functionality. But not HTML drivers API. What you are trying to do is basically 2 function calls of the existing API - `IO::NET::HTTP::REQUEST` and `PARSE`. The former returns raw body and that latter invokes a driver's parse method and returns a DOM tree.
> > 
> > 
> > These changes are not in the `HTML` driver (there is no such thing), but in the `HTTP`. In the API part, nothing changes - for the user the function remains the same, but the result may be different (for example, `xpath` can return different types and this is convenient), because the library allows you to do this, and this is great.
> 
> `HTML` driver is a general term describing an interface in the `drivers` module.
> 
> > > What you are trying to do is basically 2 function calls of the existing API - `IO::NET::HTTP::REQUEST` and `PARSE`. The former returns raw body and that latter invokes a driver's parse method and returns a DOM tree.
> > 
> > 
> > No, I'm not trying to do this - the `HTTPResponse` is returned with raw body and not the `HTMLDocument` with DOM as `Parse` does, the method does not use `Parse` because there is no need for it.
> 
> Then I do not understand why you are trying to add low-level HTTP functionality to HTML-only components.

I don't try to add low-level HTTP functionality to HTML-only components because it current base functionality for `HTTP` driver - do request to destination host and return content. Current method `Open` return content as virtual DOM. I just added the ability to return raw body, cookies and headers without build virtual DOM - this is the base and therefore my implementation does not return `HTMLDocument` but `HTTPResponse` if we don't need `HTMLDocument`.

Yes,  it will return other type but as I already gave an example - `xpath()` does the same and it's normal.

Ok, If `CDP` driver can return RAW body - ok, let's do the same, but i don't understand what case could this be.
But anyway - if someone suddenly needs it, will he also need to make his own driver?

And finally - I think `document` should have been called as just `open` and this discussion would not exist :).

> > p.s. If now I make my changes to IO::NET::HTTP, then there will be no backward compatibility. Users will have to rewrite their fqls.
> > You don't have to break backward compatibility.
> 
> Anyway, I'm not willing to accept the changes in the driver. You always can register your own driver and use it as a default one. It's one of the design goals since the library can not cover each user's edge case.

bundleman:
> Let's bring in the other changes first, and then we'll figure out what to do with the HTTP thing.

The others are not so necessary :)

PR #734, opened 2022-03-22
adding test for Options

solrac97gr:
Hi, I create the test for Options File for increase the coverage

### DIFF
--- a/pkg/runtime/options_test.go
+++ b/pkg/runtime/options_test.go
@@ -1,0 +1,42 @@
+package runtime_test
+
+import (
+	"context"
+	"io"
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+
+	"github.com/MontFerret/ferret/pkg/runtime"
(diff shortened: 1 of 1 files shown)


pkg/runtime/options_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+		options = append(options, optionTestWithParam)
+		options = append(options, optionTestWithParams)
+		options = append(options, optionWithLog)
+		options = append(options, optionWithLogLevel)
+		options = append(options, optionWithLogFields)
+
+		opts := runtime.NewOptions(options)
+
+		ctx := opts.WithContext(context.Background())
```
ziflex, line 38: I think we also need to test if all options are properly set via setters.

ziflex (changes requested):
Thanks for your PR. Please, update your unit test.

pkg/runtime/options_test.go
solrac97gr: Great, what do you mean using setters? Do you have some examples for check in the project?

Issue #606, opened 2021-03-31
Documentation: add examples of queries made using `XPATH`


ngirard:
Unless I missed something, the documentation doesn't explain how to query document metadata (searching "site:montferret.dev metadata" through Google returned nothing, neither did grepping the source code).

As an example, I tried to query the `og:url` metadata.
I tried variations of `//meta[property='og:url']::attr(content)`, with or without the leading `//`, and with or without the `attr(content)`, but was unsuccessful.

Cheers

ziflex:
Try to use [XPATH](https://www.montferret.dev/docs/stdlib/html/#xpath) function for that.

suntong:
Yeah, examples of queries made using `XPATH` would be super, whether in the doc or the cookbook.

PR #740, opened 2022-05-31
Add xpath example

slowmanchan:
Added an example for the XPATH command for [this](https://github.com/MontFerret/ferret/issues/606) issue.


ziflex:
Great, ideally we need to have it on our website too. But it's for the other day :)

Issue #489, opened 2020-04-30
Add support of synthetic navigation to HTTP driver


ziflex:
Right now, pages created by in-memory ``http`` driver do not support navigation between pages due to their nature. 
It makes it problematic to "move" from one static page to another within a same domain simply because we cannot reuse current page. In order to do so, we have to create a new page every time we need to move from A to B.
Let's add a possibility to reuse existing page by implementing synthetic navigation that would just download a new HTML page by a given URL and replace root element.

slowmanchan:
Would love to help out here but I'm not too sure whats required here. Could you please elaborate a bit more? Thanks!

slowmanchan:
Thanks for the info.
I'll check it out!

slowmanchan:
I've got a WIP [PR](https://github.com/MontFerret/ferret/pull/743). It works but is this what you mean? Also, can you let me know what your development process is? I'm currently compiling after code changes and testing manually with an `.fql` file. I'd also like to cover this with some tests but i'm not sure how to go about it (mocking drivers / objects).

PR #743, opened 2022-06-05
[WIP] Add synthetic nav to http driver

slowmanchan:
For Issue: #489

### DIFF
--- a/pkg/drivers/http/driver.go
+++ b/pkg/drivers/http/driver.go
@@ -137,7 +137,7 @@ func (drv *Driver) Open(ctx context.Context, params drivers.Params) (drivers.HTM
 		Headers:    drivers.NewHTTPHeadersWith(resp.Header),
 	}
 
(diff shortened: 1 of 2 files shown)


slowmanchan:
Stupid question but, what is the purpose of having this sort of heirarchy?

slowmanchan:
Can I get a review on this PR when you have the chance.

slowmanchan:
> Hey,
> 
> thanks for your work on this item! There is a problem with your approach but it's not your fault :)
> 
> Here is a domain hierarchy of packages:
> 
> * compiler
>   
>   * standard library
>     
>     * drivers
>       
>       * runtime
> 
> The access must be from the top to the bottom. Not vice versa. In your case, you are accessing the standard library from `http` driver which breaks the access flow rule.
> 
> What you need to do instead is reuse existing functionality in the driver itself.

slowmanchan:
@ziflex just lookin for a review of this PR when you have time!

PR #761, opened 2023-01-01
Update cdp driver to work with latest cdp protocol

WPRobson:
In commit d0c159af4b8f9066cc6c90cce5970bb0289e0c01 in https://github.com/mafredri/cdp it looks like a couple of changes were made that are not backwards compatible with the implementation in ferret so these changes should fix that.
it looks like they:
- removed the SetignoreInvalidPageRanges function
- added arguments to the dom.enable for including white space

 I got these errors immediately after cloning the repository and was unable to build/run without fixing them.

### DIFF
--- a/pkg/drivers/cdp/helpers.go
+++ b/pkg/drivers/cdp/helpers.go
@@ -3,6 +3,7 @@ package cdp
 import (
 	"context"
 	"github.com/MontFerret/ferret/pkg/runtime/events"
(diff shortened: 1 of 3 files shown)


ziflex:
Hey, thanks for your PR, but I think I fixed it accidentally in the recent updates.

WPRobson:
No worries!

PR #787, opened 2023-07-06
return all submatches in RegexMatch instead of the first

rkyleg:
test string: "39.59% annually and 0.04% monthly"

previously regex_match would only return [39.59] for a regex of "\d+[.]\d+" or [39.59,35.59] for a regex of "(\d+[.]\d+)"

this pull request fixes that and returns [39.59,0.04] for a regex of "\d+[.]\d+" and [[39.59,39.59],[0.04,0.04]] for a regex of "(\d+[.]\d+)"


ziflex:
Hey, thanks for the fix!
But, could you write some unit tests, please?

PR #1033, opened 2026-09-11
test(msgpack): use platform runtime Int bounds for range singletons

dajiaohuang:
The range encode test asserted exact `Int` round-trips of `math.MinInt64` and `math.MaxInt64`. The decoder bounds decoded integers by the native int range, so on 32-bit platforms those singletons become a `Float` or are rejected, and `TestMsgpackCodecEncode/range/maximum_singleton` fails on `GOARCH=386`:

```
codec_test.go:290: decode failed: msgpack: integer 9223372036854775807 exceeds runtime range
```

The singleton boundaries now follow the platform runtime `Int` range, so the case exercises each platform's actual boundary. 64-bit expectations are unchanged, and 32-bit boundary behavior stays covered by the existing `int_overflow_float_on_32bit` and `unsigned_overflow_error` tests.

Verified with `go test ./pkg/encoding/msgpack/` on amd64 and `GOARCH=386` (both pass). Within `./pkg/... ./uapi/... .` the 386 and amd64 failure sets are now identical.

Not addressed here: the 32-bit encoder still emits full int64 values that its decoder cannot read back, so a `MaxInt64` range round-trip fails on encode/decode asymmetry rather than in the range logic. Happy to follow up if you want that bound made symmetric.

### DIFF
--- a/pkg/encoding/msgpack/codec_test.go
+++ b/pkg/encoding/msgpack/codec_test.go
@@ -253,6 +253,11 @@ func TestMsgpackCodecEncode(t *testing.T) {
 	})
 
 	t.Run("range", func(t *testing.T) {
+		// The decoder bounds decoded integers by the native int range, so the
+		// singleton boundaries follow the runtime Int range of the platform.
+		maxRuntimeInt := int64(^uint(0) >> 1)
+		minRuntimeInt := -maxRuntimeInt - 1
+
 		tests := []struct {
 			name  string
 			want  []int64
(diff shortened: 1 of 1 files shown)


ziflex:
@dajiaohuang thank you for catching the issue!
Ferret is not officially supporting 32bit platforms due to historic reasons (in the early days of the project there were some issues compiling ANTLR parser on such platforms). 
But many things have changed since then so I'm going to revisit it.

ziflex:
@dajiaohuang could you tell me about your use case?

dajiaohuang:
The use case is cross-platform validation: Ferret's msgpack decoder converts range singleton bounds through the platform-native `int`, so the test must exercise `math.MinInt32`/`math.MaxInt32` on 32-bit builders and `int64` bounds on 64-bit builders. The production codec is unchanged; this keeps the test aligned with the runtime contract for downstream 32-bit consumers and avoids a false failure caused by impossible native-int values.

ziflex:
Fixed in #1046

